Issue #9, opened 2019-02-05
Remove calling pressHomeButton in publish


triwav:
Pressing home should no longer be needed. We should try removing it to speed up deployments and if needed making an option if it causes issues in some specific cases.

Issue #8, opened 2019-02-04
Add Signed Package Creation


triwav:
Currently roku-deploy doesn't handle anything other than doing a regular deployment to the device while developing. In order to help make this the base for a more formalized CI/CD script we need to at minimum be able to make a signed package file for upload to Roku as well. Being able to switch Dev ID would be nice as well but is not required.

TwitchBronBron:
Not sure if the "username" field is what you mean by Dev ID, but if so, then that's already supported? https://github.com/TwitchBronBron/roku-deploy/blob/d7dceeecb5d25c239b3cebf3c982dddcc5ec0eaa/src/index.ts#L599

triwav:
No referring to actually being able to change what dev ID you're signing with. You can rekey a device by providing a package that was signed with it as well at the password. Some people work on multiple projects with different dev IDs and that would simplify and hopefully also help reduce errors as well. Again just a nice to have.

TwitchBronBron:
Gotcha. Ok cool, sounds good.

PR #13, opened 2019-02-12
 #8: Add support for signed package creation, #9 stop calling home button on deploy and other good stuff

leightyb-teachco:
The main purpose of this PR centers around signed package creation from #8 but I've also added some other things like build incrementing at the same time as well as no longer calling the home button on deploy #9. I've also added testSignedPackage.pkg along with the accompanying devID and password to serve as a base for adding resigning a device with this module. I still need to write some more tests but am posting to get feedback from @TwitchBronBron to get feedback on how to go about writing those.


readme.md
```diff
@@ -56,6 +54,27 @@ From an npm script in package.json. (Requires rokudeploy.json to exist at the ro
         }
     }
 
+A callback has also been added to the higher level methods so you can modify the copied contents before the package is zipped. An info object is passed in with the following attributes
```
TwitchBronBron: Change the wording a little bit. "has been added" makes this feel like a changelog, when in reality, people are coming here to read how to do it. So something like "You can provide a callback in any of the higher level methods, which allow you do things like modify.....

readme.md
```diff
@@ -56,6 +54,27 @@ From an npm script in package.json. (Requires rokudeploy.json to exist at the ro
...
+- **manifestData:** [key: string]: string
+    Contains all the parsed values from the manifest file
+- **stagingFolderPath:** string
+    Path to staging folder for easy access to know where to modify relative to
```
TwitchBronBron: Is there more to this sentence?

readme.md
```diff
@@ -65,11 +84,14 @@ Here are the available options. The defaults are shown to the right of the optio
...
+    The password used for creating signed packages
+
 - **outDir?:** string = `"./out"`  
-    A full path to the folder where the zip package should be placed  
+    A full path to the folder where the zip/pkg package should be placed
 
-- **outFile?:** string = `"roku-deploy.zip"`  
-    The name the zip file should be given.  
+- **outFile?:** string = `"roku-deploy"`  
+    Basename the zip/pkg file should be given..
```
TwitchBronBron: Maybe say something like "The base file name that the zip/pkg should be given, without the extension."

src/index.ts
```diff
@@ -5,14 +5,16 @@ import * as globAll from 'glob-all';
...
 // tslint:disable-next-line
 export var __request: any = request;
 
 /**
  * Copies all of the referenced files to the staging folder
- * @param options 
+ * @param options
  */
-export async function prepublishToStaging(options: RokuDeployOptions) {
+export async function prepublishToStaging(options: RokuDeployOptions): Promise<string> {
```
TwitchBronBron: In general, I prefer to let typescript infer the return type. Was there a reason you specified it here?

src/index.ts
```diff
@@ -507,10 +636,65 @@ export function getOptions(options: RokuDeployOptions = {}) {
     return finalOptions;
 }
 
+export function getStagingFolderPath(options?: RokuDeployOptions): string {
```
TwitchBronBron: Remove the `: string` type here, it is inferred by typescript and is therefore not needed.

src/index.ts
```diff
@@ -507,10 +636,65 @@ export function getOptions(options: RokuDeployOptions = {}) {
...
+    stagingFolderPath = path.resolve(stagingFolderPath);
+    return stagingFolderPath;
+}
+
+
+/**
+ * Centralizes getting output zip file path based on passed in options
+ * @param options
+ */
+export function getOutputZipFilePath(options: RokuDeployOptions): string {
```
TwitchBronBron: Remove the `: string` type here; typescript can infer this from the return variable.

src/index.ts
```diff
@@ -551,13 +735,13 @@ export function zipFolder(srcFolder: string, zipFilePath: string) {
...
     /**
-     * A full path to the folder where the zip package should be placed
+     * A full path to the folder where the zip/pkg package should be placed
      * @default "./out"
      */
     outDir?: string;
     /**
-     * The name the zip file should be given. 
-     * @default "roku-deploy.zip"
+     * Basename the zip/pkg file should be given.
```
TwitchBronBron: Change to something like "Base filename the zip/pkg file should be given (excluding the extension)"

src/index.ts
```diff
@@ -602,10 +786,32 @@ export interface RokuDeployOptions {
...
     failOnCompileError?: boolean;
 }
 
-export type FilesType = (string | string[] | { src: string | string[]; dest?: string });
\ No newline at end of file
+export interface ManifestData {
+    [key: string]: string;
+}
+
+export interface RokuDeployBeforeZipCallbackInfo {
```
TwitchBronBron: Rename this to `BeforeZipCallbackInfo`. No need to prefix with `RokuDeploy`.

src/index.spec.ts
```diff
@@ -1,777 +1,898 @@
...
+                await rokuDeploy.createPackage(options);
+                assert.fail('Exception should have been thrown');
+            } catch (e) {
+                assert.ok('Exception was thrown as expected');
+            }
+        });
+
+        it('should create package in proper directory', async function () {
+            await rokuDeploy.createPackage(options);
+            expect(file(await rokuDeploy.getOutputZipFilePath(options))).to.exist;
```
TwitchBronBron: `getOutputZipFilePath` is not async, so remove the await

src/index.spec.ts
```diff
@@ -1,777 +1,898 @@
...
+        });
+
+        it('should only include the specified files', async () => {
+            try {
+                options.files = ['manifest'];
+                await rokuDeploy.createPackage(options);
+                let zip = new AdmZip(await rokuDeploy.getOutputZipFilePath(options));
```
TwitchBronBron: `getOutputZipFilePath` is not async, so remove the await

src/index.spec.ts
```diff
@@ -1,777 +1,898 @@
...
+                zip.extractAllTo('.tmp/output', true);
+                expect(file('./.tmp/output/manifest')).to.exist;
+            } catch (e) {
+                throw e;
+            }
+        });
+
+        it('generates full package with defaults', async () => {
+            await rokuDeploy.createPackage(options);
+            let zip = new AdmZip(await rokuDeploy.getOutputZipFilePath(options));
```
TwitchBronBron: `getOutputZipFilePath` is not async, so remove the await

src/index.spec.ts
```diff
@@ -1,777 +1,898 @@
...
+                expect(e.message).to.equal('Invalid response');
             }
-        ];
-        await prepublishToStaging(options);
-        expect(file('out/.roku-deploy-staging/source/renamed.brs')).to.exist;
-    });
+        });
+
+        it('should return an error if failure returned in response', async () => {
+            let body = `   <div style="display:none">
```
TwitchBronBron: Format this html a little nicer, or add a comment why it needs to be exactly this format.

```typescript
let body = `   
    <div style="display:none">
        <font color="red">Failed: Invalid Password.</font>
   </div>
`;

**Comment by TwitchBronBron** [comment]
Can you a few entries in the readme under **usage** that shows users how to use your new package creation and signing methods?

File: src/index.ts [hunk 13]
```diff
@@ -399,80 +428,179 @@ export async function pressHomeButton(host) {
...
+
+    let requestOptions = generateBaseRequestOptions('plugin_package', options);
+    requestOptions.formData = {
+        mysubmit: 'Package',
+        pkg_time: (new Date()).getTime(),
+        passwd: options.signingPassword,
+        app_name: appName,
+    };
+
+    return Promise.all([]).then(function () {
```
TwitchBronBron: You can remove this outer Promise.all. I was using that to start a promise chain certain places. However, in this instance, you start a promise chain directly underneal on the next line...so just use that.

readme.md
triwav: Makes sense. That's what happens when you're doing it at the same time you're writing commits and PRs will change in next commit.

src/index.ts
triwav: Ah! Was not aware of that. I'm just used to having to do it in Swift. Makes sense why it wasn't provided.

src/index.spec.ts
triwav: Still crazy to me that's allowed. I guess it's up to PRs to uncover this sort of stuff :)

src/index.spec.ts
triwav: I'm open to thoughts on this. I copied it exactly how it was on the device. It looks ugly in vs code since the spacing is off. I'll do a comment for now but am open to switching but didn't want the regex check to not be exactly what was happening.

TwitchBronBron:
I'm approving this. The build is failing because code coverage is below 100%. I will work on getting that back to 100%, but it's a pain.

PR #18, opened 2019-06-28
Manifest with bs_const weren't being handled correctly. This fixes it

triwav:
bs_const uses equal signs in the right hand side which ini then adds quotes around the whole thing. It doesn't seem like a fix will be available soon (https://github.com/npm/ini/issues/57) so we remove it before stringifying and add it back in afterwards to get around this.


TwitchBronBron:
Looks great!

PR #22, opened 2019-08-07
Fix manifest color

triwav:
colors starting with # symbol were being treated as comments, removed ini dependency and did simpler parsing to handle both this and bs_const issue that was previously worked around


TwitchBronBron:
There are some tslint errors that are failing the build. https://travis-ci.org/TwitchBronBron/roku-deploy/jobs/568964827#L228

PR #47, opened 2021-01-08
Ignore 202 response codes on home key presses

triwav:
Ignore 202 response codes on home key presses since this should be a fire and forget call anyways and causes issues that really aren't issues when we receive a 202 response code.


TwitchBronBron (approved):
Looks good!

PR #50, opened 2021-05-06
Added some better logic to search the response bodies for Roku messages

chrisdp:
This logic now looks at the response bodies to try and find the messages from the device as to what went wrong. It also adds a verbose log mode that for now is used to dump the device response body to the console for debugging. 

We will want to take a look at how this impacts things like rekeying since that uses an older way to detect errors from the device and this may be a braking change as a result.


src/RokuDeploy.ts
```diff
@@ -652,15 +653,54 @@ export class RokuDeploy {
...
-            throw new errors.InvalidDeviceResponseCodeError('Invalid response code: ' + results.response.statusCode);
+            throw new errors.InvalidDeviceResponseCodeError('Invalid response code: ' + results.response.statusCode, results);
         }
     }
 
+    public getRokuMessagesFromResponseBody(body: string): { errors: Array<string>; infos: Array<string>; successes: Array<string> } {
+        let errors = [];
+        let infos = [];
+        let successes = [];
+        let errorRegex = /Shell\.create\('Roku\.Message'\)\.trigger\('[\w\s]+',\s+'(\w+)'\)\.trigger\('[\w\s]+',\s+'(.*?)'\)/igm;
```
TwitchBronBron, line 677: I know this is how they are sending it currently, but what are your thoughts on matching optional whitespace around all the javascript that could change. such as:
```typescript
/Shell\s*\.\s*create\s*\(\s*'Roku\.Message'\s*\)   ......etc/
```

src/RokuDeploy.ts
```diff
@@ -652,15 +653,54 @@ export class RokuDeploy {
...
+        let match;
+
+        // eslint-disable-next-line no-cond-assign
+        while (match = errorRegex.exec(body)) {
+            let [fullMatch, messageType, message] = match;
```
TwitchBronBron, line 682: If you aren't going to use `fullMatch`, you can just use an empty comma. Like this:

```typescript
let [, messageType, message] = match;
```

src/RokuDeploy.ts
```diff
@@ -652,15 +653,54 @@ export class RokuDeploy {
...
+            switch (messageType) {
```
TwitchBronBron, line 683: cast the message type to lower case just to be safe.

src/RokuDeployOptions.ts
```diff
@@ -131,6 +131,11 @@ export interface RokuDeployOptions {
      * If true, the publish will fail on compile error
      */
     failOnCompileError?: boolean;
+
+    /**
+     * Set to true to help with debugging issues in roku-deploy
+     */
+    verboseLogging?: boolean;
```
TwitchBronBron, line 138: Let's go with `logLevel` instead. I know you only added verbose, and that's fine. I'm ok with having several of the log levels do nothing different. But I don't want to have to change this flag later on. 

```typescript
logLevel?: LogLevel


export enum LogLevel {
    error = 1,
    warn = 2,
    log = 3,
    info = 4,
    debug = 5,
    trace = 6
}
```

Feel free to copy the Logger class from brighterscript: https://github.com/rokucommunity/brighterscript/blob/master/src/Logger.ts

TwitchBronBron (changes requested):

src/RokuDeploy.ts
chrisdp: I think this is very very unlinkely. If they do make a change that breaks this it's likely not going to be white space. It's going to be function names and such.

src/RokuDeployOptions.ts
TwitchBronBron: resolved. :D

PR #55, opened 2021-05-27
Switch to using jszip for zip compression

triwav:
Switch to using jszip for zip compression which takes about one quarter of the time od archiver, add callback on zipFolder to allow us to eventually move away from needing a staging folder


.eslintignore
```diff
@@ -0,0 +1 @@
+types
```
TwitchBronBron, line 1: Instead of doing this file, let's use the `ignorePatterns` in `eslintrc.js` instead. https://eslint.org/docs/user-guide/configuring/ignoring-code#the-eslintignore-file

src/RokuDeploy.ts
```diff
@@ -938,41 +938,34 @@ export class RokuDeploy {
...
+        for (const file of files) {
+            const promise = this.fsExtra.readFile(file.src).then((data) => {
+                if (preFileZipCallback) {
+                    data = preFileZipCallback(file, data);
+                }
 
-            output.on('error', (err) => {
-                reject(err);
-            });
+                const ext = path.extname(file.dest);
```
TwitchBronBron, line 953: Should probably `.toLowerCase()` this so we can handle things like `.JPG`

Issue #64, opened 2021-09-14
Inexplicably cannot deploy if there is no currently deployed app!


georgejecook:
If no app is currently side-loaded, then we get:  

```
  Error: Delete Failed: No such file or directory.
      at new FailedDeviceResponseError (node_modules/roku-deploy/dist/Errors.js:52:28)
      at RokuDeploy.checkRequest (node_modules/roku-deploy/dist/RokuDeploy.js:873:19)
      at RokuDeploy.<anonymous> (node_modules/roku-deploy/dist/RokuDeploy.js:830:34)
      at step (node_modules/roku-deploy/dist/RokuDeploy.js:44:23)
      at Object.next (node_modules/roku-deploy/dist/RokuDeploy.js:25:53)
      at fulfilled (node_modules/roku-deploy/dist/RokuDeploy.js:16:58)
      at processTicksAndRejections (node:internal/process/task_queues:94:5)

```

this is coz the HTTP delete fails - that error should be ignored.
If I get a mo, I'll put a pr up; but it's unlikely

TwitchBronBron:
Resolved by #65

PR #65, opened 2021-09-15
wraps call to delete installed cahnnel in try catch, so that deployments continue

georgejecook:
addresses: https://github.com/rokucommunity/roku-deploy/issues/64


TwitchBronBron:
This looks good. We just need to figure out the windows build failure. I'll take a look on my windows 10 laptop soon..

TwitchBronBron (approved):

PR #79, opened 2021-11-01
run replace first, then fallback to replace.

TwitchBronBron:
Sometimes a "replace" command will fail on roku devices, so if that happens, we fall back to an "install" call which should work the rest of the time. 

I also made the `generateBaseRequestOptions` function take a formData object to simplify callsite code.

### DIFF
--- a/src/RokuDeploy.ts
+++ b/src/RokuDeploy.ts
@@ -412,7 +412,7 @@ export class RokuDeploy {
         }));
     }
 
-    private generateBaseRequestOptions(requestPath: string, options: RokuDeployOptions): request.OptionsWithUrl {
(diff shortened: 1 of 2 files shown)


src/RokuDeploy.ts
```diff
@@ -468,27 +469,37 @@ export class RokuDeploy {
...
+            });
 
+            //attach the remotedebug flag if configured
             if (options.remoteDebug) {
                 requestOptions.formData.remotedebug = '1';
             }
 
-            let results = await this.doPostRequest(requestOptions);
+            //now that the channel is deleted, try to "replace" it first since that usually works.
+            let response: HttpResponse;
```
TwitchBronBron, line 484: renamed to response to better align with what is happening.

chrisdp (approved):

PR #80, opened 2021-11-01
Use ip address for ECP press

TwitchBronBron:
ECP requests will fail if using a hostname instead of an ip address, so this will convert the hostname to an ip address only for the `pressHomeButton` call.

### DIFF
--- a/src/RokuDeploy.ts
+++ b/src/RokuDeploy.ts
@@ -435,10 +435,12 @@ export class RokuDeploy {
      * @param port - the port that should be used for the request. defaults to 8060
      * @param timeout - request timeout duration in milliseconds. defaults to 150000
      */
-    public async pressHomeButton(host, port?: number, timeout?: number) {
+    public async pressHomeButton(host: string, port?: number, timeout?: number) {
         let options = this.getOptions();
         port = port ? port : options.remotePort;
         timeout = timeout ? timeout : options.timeout;
+        //convert hostname to ip address because ECP commands only work with IP addresses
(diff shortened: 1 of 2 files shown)


TwitchBronBron:
The more I think about this, the less convinced I am that roku-deploy is where we should be doing this dns lookup. Perhaps roku-deploy should just be neutral. And then something like vscode-brightscript-language or `brighterscript` can make the decision whether it should do a dns lookup. @elsassph @georgejecook @triwav @chrisdp any thoughts?

chrisdp (approved):

src/RokuDeploy.ts
```diff
@@ -435,10 +435,12 @@ export class RokuDeploy {
      * @param port - the port that should be used for the request. defaults to 8060
      * @param timeout - request timeout duration in milliseconds. defaults to 150000
      */
-    public async pressHomeButton(host, port?: number, timeout?: number) {
+    public async pressHomeButton(host: string, port?: number, timeout?: number) {
         let options = this.getOptions();
         port = port ? port : options.remotePort;
         timeout = timeout ? timeout : options.timeout;
+        //convert hostname to ip address because ECP commands only work with IP addresses
+        host = await util.dnsLookup(host);
```
triwav, line 443: Is there not a better spot for us to put this lookup than in this single call function?

TwitchBronBron: Well, this is the only place where it would actually fail. So while we _could_ translate it other places, we don't really need to, so is there a benefit?

PR #86, opened 2022-03-15
Performance improvements

TwitchBronBron:
Fixes several performance issues related to globbing for files. 
 - run all glob lookups in parallel instead of sequentially
 - negative globs no longer run a glob and then remove those entries from the overall result. Instead, the negative globs are used to filter previous matches. This is significantly faster.
 - switch to a faster glob library. We currently use node-glob. Based on my preliminary tests, fdir is the fastest. However, there's a bug in that library right now related to symlinks, so for now we will use `fast-glob`. (EDIT: fdir has a few other design differences that prevent us from switching. fast-glob is plenty fast). 
image

To benchmark this, I ran the following roku-deploy command against a folder with >64,000 files:
```typescript
await rokuDeploy.getFilePaths([
    '**/*',
    '!**/node_modules/**/*'
], "path/to/folder");
```

Results:
**baseline (before rewrite):**
24.409 ops/sec, duration: 47s318ms

**after rewrite (using node-glob):**
916.974 ops/sec, duration: 1s626ms

**after rewrite (using fast-glob):**
3,147.138 ops/sec, duration: 0s367ms

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,8 +11,8 @@
             "dependencies": {
                 "chalk": "^2.4.2",
                 "dateformat": "^3.0.3",
+                "fast-glob": "^3.2.11",
                 "fs-extra": "^7.0.1",
-                "glob": "^7.1.6",
                 "jsonc-parser": "^2.3.0",
                 "jszip": "^3.6.0",
                 "minimatch": "^3.0.4",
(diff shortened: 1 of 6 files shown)


chrisdp (approved):

NathanGillis (approved):

livecano (approved):
Very nice!

PR #92, opened 2022-04-14
First pass at screenshot support

chrisdp:


src/RokuDeploy.ts
```diff
@@ -731,6 +716,67 @@ export class RokuDeploy {
...
+        let createScreenshotResult = await this.doPostRequest({
+            ...this.generateBaseRequestOptions('plugin_inspect', options),
+            formData: {
+                mysubmit: 'Screenshot',
+                archive: ''
+            }
+        });
+
+        // Pull the image url out of the response body
+        const imageUrlOnDevice = /["'](pkgs\/dev\.jpg\?.+?)['"]/gi.exec(createScreenshotResult.body)?.[1];
```
chrisdp, line 751: you can change the devices output from `jpg` to `png` I need to add support for this here and in the check for what format you asked for.

chrisdp: Thanks @triwav for pointing this out. 
You can do this with:
home 5x -> up -> right -> down -> left -> up
and using `Cycle screenshot output format`

TwitchBronBron:
Do you think this is a good opportunity to add another cli command to support this functionality?

Here are the options I would expect to see
```
roku-deploy screenshot --host --username --password --screenshot-path
```

TwitchBronBron (approved):

src/RokuDeploy.ts
```diff
@@ -731,6 +716,70 @@ export class RokuDeploy {
         return this.doPostRequest(deleteOptions);
     }
 
+    /**
+     * Gets a screenshot from the device. A side-loaded channel must be running or an error will be thrown.
+     */
+    public async takeScreenshot(options?: RokuDeployOptions & { screenshotPath?: string }) {
```
TwitchBronBron, line 722: Let's create a specific interface for this function that includes the options we care about (rootDir, host, password). We only use like 3 of the RokuDeployOptions.

chrisdp: Sure.

chrisdp:
> Do you think this is a good opportunity to add another cli command to support this functionality?
> 
> 
> 
> Here are the options I would expect to see
> 
> ```
> 
> roku-deploy screenshot --host --username --password --screenshot-path
> 
> ```

This would come when we add cli support. Currently we don't support this in roku-deploy.

src/RokuDeploy.ts
```diff
@@ -1051,3 +1037,28 @@ export interface HttpResponse {
...
+    host: string;
+
+    /**
+     * The password for logging in to the developer portal on the target Roku device
+     */
+    password: string;
+
+    /**
+     * A full path to the folder where the screenshots should be saved.
+     * Will use the OS temp file system if not supplied
```
TwitchBronBron, line 1055: Should this say "OS temp directory"?

src/RokuDeploy.ts
```diff
@@ -1051,3 +1037,28 @@ export interface HttpResponse {
...
+     */
+    outDir?: string;
+
+    /**
+     * The base filename the zip/pkg file should be given (excluding the extension)
```
TwitchBronBron, line 1060: This is not a zip/pkg.

src/RokuDeploy.ts
```diff
@@ -1051,3 +1037,28 @@ export interface HttpResponse {
...
+     * The default format looks something like this: screenshot-YYYY-MM-DD-HH.mm.ss.SSS.<jpg|png>
+     */
+    outFileName?: string;
```
TwitchBronBron, line 1063: I still think this should just be `outFile`. Otherwise, why not call `outDir` `outDirName`? The `dir` and `file` are the names of the things. `outFileName` tells me nothing more than `outFile`?

TwitchBronBron (changes requested):

PR #94, opened 2022-05-23
Allow filtering files during `zipFolder`

TwitchBronBron:
Adds a files array parameter to `zipFolder` to allow files to be included/excluded from the zip. This is mostly useful for vscode-brightscript-language to exclude files like sourcemaps from getting uploaded to the device (and save lots of performace doing so).


livecano (approved):
LGTM

chrisdp (approved):

PR #109, opened 2023-03-14
Use micromatch instead of picomatch

TwitchBronBron:
We use `fast-glob` for file discovery (which uses `micromatch` for glob logic), but for some reason were using `picomatch` for our path filtering logic. 

This PR aligns on `micromatch` for path filtering logic, meaning that now all glob logic within roku-deploy is handled using the same library.


TwitchBronBron:
> But is that the same version as `fast-glob` is using otherwise that would be duplicated anyway?

Good call! There was a patch release of fast-glob (`3.2.11` -> `3.2.12`), so I upgraded that, and then aligned our version of micromatch with the one from fast-glob (`4.0.4`)

PR #115, opened 2023-05-09
TBD-67: roku-deploy: fix nodejs 19 bug

fumer-fubotv:
'Request' module post requests fail with Nodejs v19 and up. To make it compatible with newer versions of node simply pass the following config values to the request apis to overcome socket error: agentOptions: { 'keepAlive': false }


TwitchBronBron:
Looks like it works for me on node 16 and node 19.

TwitchBronBron (approved):

PR #120, opened 2023-11-02
Enhance getDeviceInfo() method

TwitchBronBron:
Enhances the `getDeviceInfo()` method:
- add `enhance` option to get an object back in `camelCase` instead of the standard `kebab-case`, and to convert boolean strings to booleans and number strings to numbers
- retains backwards compatibility by adding an overloaded signature that is not the default


src/util.spec.ts
```diff
@@ -3,10 +3,18 @@ import { expect } from 'chai';
...
 import { tempDir } from './testUtils.spec';
 import * as path from 'path';
+import * as dns from 'dns';
+import { createSandbox } from 'sinon';
+const sinon = createSandbox();
 
 describe('util', () => {
     beforeEach(() => {
         fsExtra.emptyDirSync(tempDir);
+        sinon.restore();
```
triwav, line 13: We want to call restore in both before and after?

TwitchBronBron: I'm sure one or the other is _probably_ fine? It's a habit I've started. You never know if the last test cleaned itself up properly, so since it's just a line or two, I just do it both places.

triwav (approved):

src/RokuDeploy.ts
```diff
@@ -919,26 +923,74 @@ export class RokuDeploy {
...
 
-    public async getDeviceInfo(options?: RokuDeployOptions) {
-        options = this.getOptions(options);
+    /**
+     * Get the `device-info` response from a Roku device
+     * @param host the host or IP address of the Roku
+     * @param port the port to use for the ECP request (defaults to 8060)
+     */
+    public async getDeviceInfo(options?: { enhance: true } & GetDeviceInfoOptions): Promise<DeviceInfo>;
+    public async getDeviceInfo(options?: GetDeviceInfoOptions): Promise<DeviceInfoRaw>
```
triwav, line 932: Do you need a semicolon here?

TwitchBronBron: Yeah it's required. it's part of the syntax requirements for overloaded method signatures. 

https://www.tutorialsteacher.com/typescript/function-overloading

PR #131, opened 2023-11-29
Don't normalize file patterns

MilapNaik:
Remove changing backward slashes to forward slashes


TwitchBronBron:
@MilapNaik looking good so far. Can you update the README to explain that the files array string and `{ src }` patterns are required to follow glob style (i.e. forward-slashes only). You can link to this writeup so we don't have to reproduce most of the documentation. 

https://www.npmjs.com/package/fast-glob?activeTab=readme#how-to-write-patterns-on-windows

README.md
```diff
@@ -256,8 +256,7 @@ The object structure is as follows:
 }
 ```
 #### { src; dest } Object Rules
- - if `src` is a non-glob path to a single file, then `dest` should include the filename and extension. For example:
- `{ src: "lib/Promise/promise.brs", dest: "source/promise.brs"}`
+ - if `src` is required to follow glob style (i.e. forward-slashes only). You can learn more [here](https://www.npmjs.com/package/fast-glob?activeTab=readme#how-to-write-patterns-on-windows).
```
TwitchBronBron, line 259: I think the original message was actually correct, and gets lost with the new message. The point of this comment is "if you reference 1 file in src, please reference 1 file in dest". 

We should add a 
```suggestion
- if `src` is a non-glob path to a single file, then `dest` should include the filename and extension. For example:
```

README.md
```diff
@@ -326,7 +325,7 @@ Here are the available options. The defaults are shown to the right of the optio
         "manifest"
     ]
     ```
-    An array of file paths, globs, or {src:string;dest:string} objects that will be copied into the deployment package.
+    An array of file paths, globs, or {src:string;dest:string} objects that will be copied into the deployment package. This is required to use glob style. You can learn more [here](https://www.npmjs.com/package/fast-glob?activeTab=readme#how-to-write-patterns-on-windows).
```
TwitchBronBron, line 328:
```suggestion
    An array of file paths, globs, or `{ src: string; dest: string }` objects that will be copied into the deployment package. Make sure to _exclusively_ use forward slashes ( `/` ) for path separators (even on Windows), as backslashes are reserved for character escaping. You can learn more about this requirement [here](https://www.npmjs.com/package/fast-glob?activeTab=readme#how-to-write-patterns-on-windows).
```

TwitchBronBron (changes requested):
just a few docs changes. everything else looks good.

TwitchBronBron (approved):
looks great!

Issue #85, opened 2022-03-14
Bug with {src;dest} object handling


TwitchBronBron:
I would expect these two patterns to be equivalent:
```
"components/*",
{ src: "components/*" }
```

According to [the documentation](https://github.com/rokucommunity/roku-deploy#-src-dest--object-rules), for `{src;dest;}` objects, when `dest` is omitted, `the root of the output folder is assumed`. However, the current implementation (**v3.5.3**) does not work that way. 

Consider the file `components/CustomButton.brs`. 
- `"components/*"` results in `pkg:/components/CustomButton.brs`.  (as expected)
- `{src: "components/*" }` results in `pkg:/CustomButton.brs` (unexpected)

This should be corrected so that the `src;dest` pattern works the same as the raw string approach. However, this would be a breaking change, so we need to consider that.

PR #135, opened 2024-01-02
Bug with {src;dest} object handling

MilapNaik:
Make change to assume dest folder. Delete 2 old tests that looked for old behavior, add 2 more tests that look for new behavior

https://github.com/rokucommunity/roku-deploy/issues/85

### DIFF
--- a/src/RokuDeploy.ts
+++ b/src/RokuDeploy.ts
@@ -329,9 +329,11 @@ export class RokuDeploy {
             //`pattern` is some other glob magic
         } else {
             const fileNameAndExtension = path.basename(srcPath);
-            const foundFilePath = util.standardizePath(`${entry.dest ?? ''}/${fileNameAndExtension}`);
-            const assumedFilePath = util.stringReplaceInsensitive(srcPath, rootDir, '');
-            result = entry.dest ? foundFilePath : assumedFilePath;
+            if (entry.dest) {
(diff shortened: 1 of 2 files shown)


src/RokuDeploy.ts
```diff
@@ -329,7 +329,9 @@ export class RokuDeploy {
             //`pattern` is some other glob magic
         } else {
             const fileNameAndExtension = path.basename(srcPath);
-            result = util.standardizePath(`${entry.dest ?? ''}/${fileNameAndExtension}`);
+            const foundFilePath = util.standardizePath(`${entry.dest ?? ''}/${fileNameAndExtension}`);
+            const assumedFilePath = util.stringReplaceInsensitive(srcPath, rootDir, '');
+            result = entry.dest ? foundFilePath : assumedFilePath;
```
TwitchBronBron, line 334: This will result in evaluating both foundFilepath and assumedFilepath every time, which is a little CPU wasteful. Let's do an `if` statement or a ternary expression instead:

```suggestion
            if (entry.dest) {
                result = util.standardizePath(`${entry.dest ?? ''}/${fileNameAndExtension}`);
            } else {
                result = util.stringReplaceInsensitive(srcPath, rootDir, '');
            }
```

TwitchBronBron (approved):

PR #142, opened 2024-02-02
Add cli commands and rename roku-deploy functions, reorganize functions

### LINKED ISSUE #70: CLI commands for various roku-deploy actions
We should be able to do everything in roku-deploy from the command-line. Things like move-to-stage, create zip from a folder, etc. 

Here's a good example of a cli file that this feature can be modelled after. 

https://github.com/rokucommunity/ropm/blob/master/src/cli.ts

We want to expose most of the public functions on the [RokuDeploy](https://github.com/rokucommunity/roku-deploy/blob/master/src/RokuDeploy.ts) class. 

So tasks for this:
 - install yargs and add distinct cli commands for all the important RokuDeploy functions, with their args
 - make dedicated commands/ folder with a file per command (like in ropm https://github.com/rokucommunity/ropm/tree/master/src/commands)

Here are a few examples of how we'd use it (args and names might need to be changed, but this gets the point across):
```bash
npx roku-deploy copy-to-staging --stagingDir ./staging
npx roku-deploy create-zip --stagingDir ./staging --file app.zip
npx roku-deploy deploy --host 192.168.1.123 --password --file app.zip
```

### LINKED ISSUE #138: Remove rokudeploy.json and bsconfig.json loading from rokudeploy functions.
Once #70 lands, we need to move the json file loading OUT of the inner rokuDeploy functions and into the cli instead. The CLI should load the json files in the same way, and augment the options object with the data found in the json files (cli args override json obj props).

TwitchBronBron:
Adds all method and CLI renames we want to do. This thing is totally broken though, needs lots of love. :D 

Fixes #70
Fixes #138


package.json
```diff
@@ -101,7 +101,7 @@
         ],
         "sourceMap": true,
         "instrument": true,
-        "check-coverage": true,
+        "check-coverage": false,
```
TwitchBronBron, line 104: We'll re-enable this once most of v4 is finished. No need to chase 100% right now

src/RokuDeploy.spec.ts
```diff
@@ -699,182 +637,6 @@ describe('index', () => {
 
     });
 
-    describe('createPackage', () => {
```
TwitchBronBron, line 702: I think we should bring back most of these tests. Looks like maybe we just need to tweak some of them to test against the new rokuDeploy function names?

src/RokuDeploy.spec.ts
```diff
@@ -2169,96 +1938,30 @@ describe('index', () => {
...
-            expect(() => rokuDeploy['normalizeFilesArray'](<any>[new Date()])).to.throw();
-            expect(() => rokuDeploy['normalizeFilesArray'](<any>[1])).to.throw();
-            expect(() => rokuDeploy['normalizeFilesArray'](<any>[{ src: true }])).to.throw();
-            expect(() => rokuDeploy['normalizeFilesArray'](<any>[{ src: /asdf/ }])).to.throw();
-            expect(() => rokuDeploy['normalizeFilesArray'](<any>[{ src: new Date() }])).to.throw();
-            expect(() => rokuDeploy['normalizeFilesArray'](<any>[{ src: 1 }])).to.throw();
-        });
-    });
-
-    describe('deploy', () => {
```
TwitchBronBron, line 2183: Let's try and re-enable these tests...might need to move them to test the `ExecCommand` instead if that's where the full flow happens?

src/RokuDeploy.spec.ts
```diff
@@ -3291,231 +2933,41 @@ describe('index', () => {
...
-        });
-
-        it('handles non-falsey values', () => {
-            expect(rokuDeploy.normalizeRootDir(cwd)).to.equal(cwd);
-            expect(rokuDeploy.normalizeRootDir('./')).to.equal(cwd);
-            expect(rokuDeploy.normalizeRootDir('./testProject')).to.equal(path.join(cwd, 'testProject'));
-        });
-    });
-
-    describe('retrieveSignedPackage', () => {
```
TwitchBronBron, line 3417: I think these tests need to be restored as well

src/RokuDeploy.spec.ts
```diff
@@ -3748,85 +3090,25 @@ describe('index', () => {
...
             const writeStream = await writeStreamPromise;
             writeStream.write('test');
             writeStream.close();
 
             await finalPromise;
-            expect(getToFileIsResolved).to.be.true;
-        });
-    });
-
-    describe('deployAndSignPackage', () => {
```
TwitchBronBron, line 3773: Restore these tests. Might need to test the ExecCommand to test these flows

src/RokuDeploy.ts
```diff
@@ -581,6 +422,11 @@ export class RokuDeploy {
         let parsedManifest = await this.parseManifest(manifestPath);
         let appName = parsedManifest.title + '/' + parsedManifest.major_version + '.' + parsedManifest.minor_version;
 
+        //prevent devId mismatch (if devId is specified)
+        if (options.devId && options.devId !== await this.getDevId()) {
+            throw new Error('devId mismatch. nope, not gonna sign');
```
TwitchBronBron, line 427: Let's make a better error message here. Maybe something like:
```
`Package signing cancelled: devId '${options.devId}' does not match on-device devId `${onDeviceDevId}`
```

src/RokuDeploy.ts
```diff
@@ -1059,21 +824,25 @@ export class RokuDeploy {
         }
     }
 
+    /**
+     * @param options
```
TwitchBronBron, line 828: Need a better description here

src/cli.spec.ts
```diff
@@ -163,58 +160,12 @@ describe('cli', () => {
         });
     });
 
-    it('Retrieves a signed package', async () => {
```
TwitchBronBron, line 166: Can we restore these tests and make them work with how the CLI runs? Or move them to a different file and test the ExecCommand instead?

src/cli.ts
```diff
@@ -1,160 +1,243 @@
...
     }, (args: any) => {
-        return new PublishCommand().run(args);
+        return new ExecCommand(
+            'close|rekey|stage|zip|delete|close|sideload|squash|sign',
+            args
+        ).run();
+    })
+
+    .command('exec', 'larger command for handling a series of smaller commands', () => {
+        return attachCommonArgs;
```
TwitchBronBron, line 83: Let's just go ahead and inline the options here.

TwitchBronBron, line 83: I think this command needs a little more description. Maybe even a few examples using .example() (or whatever yargs provides for showing examples)

src/device.spec.ts
```diff
@@ -62,31 +61,5 @@ describe('device', function device() {
...
-    describe('deployAndSignPackage', () => {
-        it('works', async () => {
-            await rokuDeploy.deleteInstalledChannel(options as any);
-            await rokuDeploy.rekeyDevice(options as any);
-            expectPathExists(
-                await rokuDeploy.deployAndSignPackage(options as any)
-            );
-        });
-    });
+    console.log(options); // So there are no errors about unused variable
```
TwitchBronBron, line 64: If we're not actually using options, let's delete it. Either way, delete this comment.

TwitchBronBron (commented):
Left a few comments. Looking great!

src/RokuDeploy.spec.ts
MilapNaik: Done

src/cli.spec.ts
MilapNaik: Decided not to reimplement

src/RokuDeploy.ts
MilapNaik: Done and added test for it

src/device.spec.ts
MilapNaik: done

src/RokuDeploy.ts
MilapNaik: Done

src/cli.ts
MilapNaik: Done

MilapNaik: Done

src/RokuDeploy.spec.ts
MilapNaik: Done

src/RokuDeploy.spec.ts
MilapNaik: Done

src/RokuDeploy.spec.ts
MilapNaik: Done

src/RokuDeploy.ts
```diff
@@ -1284,6 +994,7 @@ export interface PublishOptions {
     retainDeploymentArchive?: boolean;
```
TwitchBronBron, line 994: Missing `outFile?: string;`

src/RokuDeploy.ts
```diff
@@ -1253,29 +939,53 @@ export interface GetDeviceInfoOptions {
...
+
+export interface SendTextOptions extends SendKeyEventOptions {
+    action?: 'keypress';
+    text: string;
 }
 
-export interface CreatePackageOptions {
+export interface CloseChannelOptions {
+    host: string;
+    remotePort: number;
```
TwitchBronBron, line 973:
```suggestion
    remotePort?: number;
```

src/cli.ts
```diff
@@ -1,160 +1,242 @@
...
+import { KeyDownCommand } from './commands/KeyDownCommand';
 
 void yargs
-    .command('prepublishToStaging', 'Copies all of the referenced files to the staging folder', (builder) => {
+
+    .command('bundle', 'execute build actions for bundling app', (builder) => {
         return builder
-            .option('stagingDir', { type: 'string', description: 'The selected staging folder', demandOption: false })
-            .option('rootDir', { type: 'string', description: 'The selected root folder to be copied', demandOption: false });
+            .option('rootDir', { type: 'string', description: 'The selected root folder to be copied', demandOption: false })
```
TwitchBronBron, line 24: Add missing `outFile` option

PR #145, opened 2024-02-08
Retry the convertToSquahsfs request given the HPE_INVALID_CONSTANT error

Christian-Holbrook:
We occasionally see the convertToSquashfs request fail. It seems determined by the size of the zip and the length of the file name. The Roku responds with half of the response body and roku-deploy cannot handle the bad data.

The squashfs conversion looks like it successfully created, the error is in the response.

The work around is to make a second request to squashfs when we see the 'HPE_INVALID_CONSTANT' error. If that response includes 'fileType': 'squashfs' then we return the result saying it succeeded.

### DIFF
--- a/src/RokuDeploy.ts
+++ b/src/RokuDeploy.ts
@@ -515,7 +515,7 @@ export class RokuDeploy {
             //wrong combination. The device fails to respond to our request with a valid response.
             //The device successfully converted the zip, so ping the device and and check the response
             //for "fileType": "squashfs" then return a happy response, otherwise throw the original error
-            if ((error as any)?.code === 'HPE_INVALID_CONSTANT') {
+            if ((error as any).code === 'HPE_INVALID_CONSTANT') {
                 try {
                     results = await this.doPostRequest(requestOptions, false);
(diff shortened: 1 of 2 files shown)


src/RokuDeploy.ts
```diff
@@ -507,8 +507,23 @@ export class RokuDeploy {
             archive: '',
             mysubmit: 'Convert to squashfs'
         });
-
-        let results = await this.doPostRequest(requestOptions);
+        let results;
+        try {
+            results = await this.doPostRequest(requestOptions);
+        } catch (error) {
+            if ((error as any)?.code === 'HPE_INVALID_CONSTANT') {
```
TwitchBronBron, line 514: We could use a comment here explaining what we're doing. Probably a more detailed comment would be best, cuz this is a weird one. :D

Christian-Holbrook: 👍

Christian-Holbrook: My comment is a bit wordy. Probably want to nitpick it a bit.

TwitchBronBron (approved):

Issue #149, opened 2024-03-04
Throw exceptions inside RokuDeploy.ts when required parameters are missing


TwitchBronBron:
When calling `RokuDeploy.<name of function>`, any required options that were not provided by the options parameter, and that were not filled in by default values, should cause an exception to be thrown. 

For example, calling `RokuDeploy.captureScreenshot()` should throw an exception if we don't have `host` or `password`.

PR #156, opened 2024-03-23
Throw exceptions on missing options

MilapNaik:
Fixes https://github.com/rokucommunity/roku-deploy/issues/149


src/RokuDeploy.ts
```diff
@@ -237,9 +235,10 @@ export class RokuDeploy {
      * @param options
      */
     public async sideload(options: SideloadOptions): Promise<{ message: string; results: any }> {
+        this.checkRequiredOptions(options, ['host', 'password']);
         options = this.getOptions(options) as any;
         if (!options.host) {
-            throw new errors.MissingRequiredOptionError('must specify the host for the Roku device');
+            throw new Error('must specify the host for the Roku device');
```
TwitchBronBron, line 241: This is handled by checkRequiredOptions, right? I see several other places in this file that do this check. I think those can all be removed in favor of checkRequiredOptions.

TwitchBronBron (changes requested):
Just one suggestion.

src/RokuDeploy.ts
MilapNaik: Oh oops I meant to do this. Will remove

MilapNaik: Fixed

TwitchBronBron (approved):

PR #169, opened 2024-05-28
Add interactive remote mode

triwav:


src/cli.ts
```diff
@@ -148,6 +149,14 @@ void yargs
         return new SendTextCommand().run(args);
     })
 
+    .command('interactive', 'Provides a way to send a series of ECP key events similar to how Roku Remote Tool works but from the command line', (builder) => {
```
TwitchBronBron, line 152: Chris and I played around with some alternative names. How about:

`remote-control` 

So usage would be:

```
npx roku-deploy remote-control
```

We could add an alias like `RC` or `R` or something if that feels verbose?

triwav: updated. Added `rc` shorter version as well

src/commands/RemoteControl.ts
```diff
@@ -0,0 +1,104 @@
+import * as readline from 'readline';
+import type { RokuKey } from '../index';
+import { rokuDeploy, util } from '../index';
+
+export class RemoteControl {
```
TwitchBronBron, line 5: Let's call this class `RemoteControlCommand` and the file `RemoteControlCommand.ts` to align with the other files in this folder.

TwitchBronBron (approved):
Once you rename the class, then this can be merged. Nice work!

PR #176, opened 2024-10-06
fixes #175 - updated regex to find a signed package on `/plugin_package` page

### LINKED ISSUE #175: `signExistingPackage` returns incorrect package's remote path
when I use a device with sdcard installed `signExistingPackage` returns incorrect remote path.

For example `/plugin_package` source: https://gist.github.com/7thsky/9654b9cae6443243f540434110dd3814
[The current implementation](https://github.com/rokucommunity/roku-deploy/blob/master/src/RokuDeploy.ts#L626-L629) tries to find `<a />` link in the page - but that link is incorrect.

the `/plugin_package` page is always has `var params = JSON.parse('<json-params>')` line. In case where the device has sdcard installed that json will look like, and `json.packages[0].pkgPath` value will is the correct package's remote path.

```
  "messages": null,
  "metadata": {
    "dev_id": "<dev_id>",
    "dev_key": true,
    "voice_sdk": false
  },
  "packages": [
    {
      "appType": "channel",
      "archiveFileName": "some-package.zip",
      "fileType": "zip",
      "id": "0",
      "location": "sdcard",
      "md5": "da4a98f08d45aea6e14a481ff481ffbe",
      "pkgPath": "pkgs/sdcard0/Pae6cec1eab06a45ca1a7f5b69edd3a20.pkg",
      "size": "455694"
    }
  ]
}
```

7thsky:
fixes #175 

the current implementation tries to parse a package's remote path from  `<a .../>` on `/plugin_package`. that link is not always correct ( for example if you device has a sdcard installed )

`/plugin_package` page response always have `var params = JSON.parse('......')` line. And the actual link ( when you opens the page in the browser ) built from the following js 

```javascript
var pkgs = params.packages.filter(p => p.pkgPath !== "");
var hasPkgApp = pkgs.length > 0;
{
  var pkgDiv = document.createElement('div');
  if (hasPkgApp) {
    var package = pkgs[0];
    var pkgPath = package.pkgPath;
    var pkgName = package.pkgPath.substr(package.pkgPath.lastIndexOf('/') + 1);
    var size = package.size;

    pkgDiv.innerHTML = `<label>Currently Packaged Application:</label>
      <div>
        <font face="Courier">
          <a href="${pkgPath}">${pkgName}</a>
          <br> package file (${size} bytes)
        </font>
      </div>`;
  }
  else {
    pkgDiv.innerHTML = `<label>Currently Packaged Application:</label>
      <div>
        <font face="Courier">
          Application is not yet packaged
        </font>
      </div>`;
  }
}
```

### DIFF
--- a/src/RokuDeploy.spec.ts
+++ b/src/RokuDeploy.spec.ts
@@ -1473,13 +1473,10 @@ describe('index', () => {
         });
 
         it('should return created pkg on success', async () => {
-            let body = `var pkgDiv = document.createElement('div');
-                        pkgDiv.innerHTML = '<label>Currently Packaged Application:</label><div><font face="Courier"><a href="pkgs//P6953175d5df120c0069c53de12515b9a.pkg">P6953175d5df120c0069c53de12515b9a.pkg</a> <br> package file (7360 bytes)</font></div>';
-                        node.appendChild(pkgDiv);`;
-            mockDoPostRequest(body);
+            mockDoPostRequest(fakePluginPackageResponse);
 
             let pkgPath = await rokuDeploy.signExistingPackage(options);
-            expect(pkgPath).to.equal('pkgs//P6953175d5df120c0069c53de12515b9a.pkg');
+            expect(pkgPath).to.equal('pkgs/sdcard0/Pae6cec1eab06a45ca1a7f5b69edd3a20.pkg');
         });
 
         it('should return our fallback error if neither error or package link was detected', async () => {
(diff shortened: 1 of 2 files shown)


TwitchBronBron (approved):
Hey, this is great! Thanks for your contribution. 

I pushed a small change that includes a test for the old logic, and added the old logic back in as a fallback if the json approach doesn't work. Since this is a fairly significant change (logically) from how we derive this value, I'm hesitant to 100% switch to the new way, so having the fallback feels like a good compromise.

Do you mind pulling this version down and testing it? Once you're good with my changes, I'm happy to merge and cut a release.

7thsky:
@TwitchBronBron yes, a fallback is a good idea.

PR #180, opened 2024-11-20
Identify when a 577 error is thrown, send a new developer friendly message

Christian-Holbrook:


src/Errors.ts
```diff
@@ -55,3 +55,16 @@ export class MissingRequiredOptionError extends Error {
...
 }
+
+export class DeviceFailedConnectionError extends Error {
+    results: any;
+
+    constructor() {
+        super();
+        this.message = `Your device needs to check for updates before accepting connections. Please navigate to System Settings and check for updates and then try again.
+
+https://support.roku.com/article/208755668.`;
```
TwitchBronBron, line 66: The template string here is kinda weird how it's capturing newlines and not indented. Let's just insert newlines ourselves. 
```suggestion
        this.message = `Your device needs to check for updates before accepting connections. Please navigate to System Settings and check for updates and then try again.\n\nhttps://support.roku.com/article/208755668.`;
```

TwitchBronBron (approved):
I left one suggestion, but then this looks good to go!

src/RokuDeploy.ts
```diff
@@ -468,8 +468,16 @@ export class RokuDeploy {
...
                     throw new errors.CompileError('Compile error', replaceError, replaceError.results);
                 } else {
-                    requestOptions.formData.mysubmit = 'Install';
-                    response = await this.doPostRequest(requestOptions);
+                    try {
+                        response = await this.doPostRequest(requestOptions);
+                    } catch (installError: any) {
+                        switch (installError.results.response.statusCode) {
+                            case 577:
+                                throw new errors.DeviceFailedConnectionError();
```
TwitchBronBron, line 476: Pass in the `installError` and set it as `.cause` on the new error

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause

TwitchBronBron (approved):
Once the comments are addressed, I'm good with this!

PR #237, opened 2026-03-22
Expanded the default files array to include the locale folder

chrisdp:


chrisdp:
@copilot We want to change this PR to just add locale to the files array. Lets remove the other entries we added. Also update with mater and fix the conflicts.

src/RokuDeployOptions.ts
```diff
@@ -38,6 +38,8 @@ export interface RokuDeployOptions {
             'source/**\/*.*',
             'components/**\/*.*',
             'images/**\/*.*',
+            'fonts/**\/*.*',
```
chrisdp:
```suggestion
```

TwitchBronBron (approved):

PR #243, opened 2026-03-26
Download actual signed package from device on newer firmware

MilapNaik:
PR to update v4.0.0-alpha.2 tag with changes for #239


MilapNaik:
Moved to #244

PR #244, opened 2026-03-26
Download actual signed package from device on newer firmware

MilapNaik:
Moving fixes in #243 to here. Fixes #239.

### DIFF
--- a/src/RokuDeploy.ts
+++ b/src/RokuDeploy.ts
@@ -512,15 +512,13 @@ export class RokuDeploy {
 
         //grab the package url from the JSON on the page if it exists (https://regex101.com/r/1HUXgk/1)
         let pkgSearchMatches = /"pkgPath"\s*:\s*"(.*?)"/.exec(results.body);
-        if (pkgSearchMatches) {
-            return pkgSearchMatches[1];
+        if (!pkgSearchMatches) {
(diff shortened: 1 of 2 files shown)


src/RokuDeploy.ts
```diff
@@ -512,15 +512,13 @@ export class RokuDeploy {
 
         //grab the package url from the JSON on the page if it exists (https://regex101.com/r/1HUXgk/1)
         let pkgSearchMatches = /"pkgPath"\s*:\s*"(.*?)"/.exec(results.body);
-        if (pkgSearchMatches) {
-            return pkgSearchMatches[1];
+        if (!pkgSearchMatches) {
+            //for some reason we couldn't find the pkgPath from json, look in the <a> tag
+            pkgSearchMatches = /<a href="(pkgs\/[^\.]+\.pkg)">/.exec(results.body);
```
TwitchBronBron, line 517: Can you add a unit test with a mock response with this other style of response which doesn't have the pkgPath in json?

MilapNaik: Added a test so that both ways are tested. Technically it was the JSON path that was untested before

TwitchBronBron (approved):

TwitchBronBron (approved):

PR #248, opened 2026-04-10
Fix path.resolve() discarding staging prefix for leading-slash dest

cewert:
PR #245 replaced string interpolation with `path.resolve()` in `copyToStaging()` and `zipFolder()`. When `dest` starts with `/`, `path.resolve()` silently discards the base staging path, causing `EACCES: permission denied, mkdir '/source'`.

## Changes

- Replace `path.resolve()` in `copyToStaging()` with defensive logic that checks if `dest` is already an absolute path under `stagingPath` (use as-is) or otherwise strips leading slashes and uses `path.join()`
- Apply the same pattern in `zipFolder()` for zip entry path construction
- Add 5 regression tests covering leading-slash dest values in both functions (stubbed and integration)
- Maintain 100% code coverage

### DIFF
--- a/src/RokuDeploy.spec.ts
+++ b/src/RokuDeploy.spec.ts
@@ -828,6 +828,41 @@ describe('RokuDeploy', () => {
 
             expect(copyPaths[0].dest).to.equal(s`${stagingDir}`);
         });
+
+        it('handles dest with leading slash without escaping staging dir', async () => {
+            const copyPaths = [] as Array<{ src: string; dest: string }>;
+            sinon.stub(rokuDeploy.fsExtra, 'ensureDir').returns(Promise.resolve() as any);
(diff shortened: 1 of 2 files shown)


TwitchBronBron:
Hey, thanks for this. I think the bug in the VS Code extension was caused by something slightly different, so I’m not sure this is the implementation I want to add. The absolute change made in #245 supports writing things outside of outDir and rootDir, which was necessary for when vscode users want to set `outDir` somewhere above the workspace folder. I think this PR would introduce some additional restrictions that we don't actually need. 

I just opened https://github.com/rokucommunity/roku-deploy/pull/249 and https://github.com/rokucommunity/roku-debug/pull/329 to resolve the vscode issue.

PR #281, opened 2026-06-02
Merge master into v4

Christian-Holbrook:
## Summary                                               
  - Port `validateDeveloperPassword()` function for validating device
  credentials                                                                  
  - Port `stringifyManifest()` function to convert ManifestData back to string
  format                                                                       
  - Add `autoLaunch` option support in sideload to control channel auto-launch
  behavior                               
  - Port glob negation prefix fix (`standardizeSrcPattern`) to preserve `!`
  patterns
  - Include case-insensitive filesystem glob matching fix


TwitchBronBron (approved):

src/RokuDeploy.ts
```diff
@@ -1096,6 +1138,34 @@ export class RokuDeploy {
         return manifestData;
     }
 
+    /**
+     * Convert a ManifestData object back into a manifest file string.
+     * If keyIndexes and lineCount are present, preserves original line ordering.
+     */
+    public stringifyManifest(manifestData: ManifestData): string {
```
TwitchBronBron: I think maybe let's make this private so it's not part of the public API

src/util.ts
```diff
@@ -174,14 +180,42 @@ export class Util {
         return matchesByIndex;
     }
 
+    private async getIsFileSystemCaseSensitive(cwd: string) {
```
TwitchBronBron, line 190: There are a few places in roku-deploy v3 that leverage this now. are those reflected in the merge? (might be worth looking at https://github.com/rokucommunity/roku-deploy/pull/279 to see what items are passing this in)

TwitchBronBron (changes requested):

src/RokuDeploy.ts
Christian-Holbrook: Removed this function.

PR #282, opened 2026-06-03
Migrate networking library from `postman-request` to `needle`

### LINKED ISSUE #267: Drop postman-request in favor of native fetch
Let's migrate to using native fetch for this system. That will require a minimum of Node.js v18, but that's okay. this is a major rewrite. 

You can leverage the existing examples that we added in https://github.com/rokucommunity/roku-deploy/pull/252 and build upon that.

TwitchBronBron:
## What

Replaces the unmaintained [`postman-request`](https://www.npmjs.com/package/postman-request) with [`needle`](https://www.npmjs.com/package/needle).

**Non-breaking.** roku-deploy attaches a `request`-shaped `results`/`response` object to thrown errors and return values, so a thin shim ([`src/request.ts`](src/request.ts)) reconstructs that exact shape on top of needle — `response.statusCode`, `response.headers`, `response.request.host`, and a string `body`.

Also drops ~860 lines from the lockfile.

## How

The shim exposes the slice of the `request` API roku-deploy uses — `post(opts, cb)`, `get(opts, cb)`, and the callback-less streaming `get(opts).on(...).pipe(...)` — and translates it to needle. Digest auth is now native to needle (`auth: 'digest'`), matching postman-request's challenge/response behavior.

Four quirks surfaced while validating against a real Roku, each handled in the shim:

- **Empty multipart values** — `request` silently dropped empty fields; needle throws `"value missing for multipart!"`. The shim drops them.
- **Bodyless POSTs** — needle throws `"Empty multipart body"` on an empty body (e.g. ECP keypress in `pressHomeButton`), so `multipart` is only enabled when form data is actually present.
- **Buffer body** — under `parse_response: false` needle returns a `Buffer`; `request` returned a string and `checkRequest` guards on `typeof body === 'string'`. The shim coerces to a string.
- **Streaming digest dance** — needle emits an intermediate `401` response before retrying with `Authorization`; `getToFile` treats any non-200 as failure. The shim swallows that one intermediate 401.

## Testing

- **Unit suite:** all passing; spec stubs target `needle` directly.
- **Real device:** [`src/device.spec.ts`](src/device.spec.ts) exercises every networking-enabled `RokuDeploy` method against a real Roku, with type/shape-tolerant assertions so it won't break as device data changes. Signing and disruptive (reboot / update-check) tests are gated behind `ROKU_SIGNING_PASSWORD` / `ROKU_RUN_DISRUPTIVE=1`.

Validated end-to-end against a device: `getDeviceInfo`, `getDevId`, `getEcpNetworkAccessMode`, `validateDeveloperPassword`, `pressHomeButton`, `deploy`/`publish` (digest + multipart zip upload), `takeScreenshot` (POST + stream-to-file), `convertToSquashfs`, `deleteInstalledChannel`, `deleteComponentLibrary`/`deleteAllComponentLibraries`, and the firmware-gated `rebootDevice`/`checkForUpdate`.


src/device.spec.ts
chrisdp, line 1: lets just pull the unrelated test changes into it's own PR with justifications. Feels like we have some existing tests changed that didn't need changing. Love the new tests though.

TwitchBronBron: Done

chrisdp (approved):

PR #284, opened 2026-06-04
Remove getOptions

### LINKED ISSUE #276: Eliminate `RokuDeploy.getOptions()` in favor of per-function option sanitizing
`RokuDeploy.getOptions()` is too massive, it does too many things. We need each function to simply do their own focused bit of sanitization instead. It might be a little more duplication, but it's targeted at each function.

Christian-Holbrook:
Each function now sets its own defaults instead of relying on a centralized getOptions() function. Fixes #276

### DIFF
--- a/src/RokuDeploy.spec.ts
+++ b/src/RokuDeploy.spec.ts
@@ -4293,6 +4293,139 @@ describe('RokuDeploy', () => {
         });
     }
 
+    describe('defaults', () => {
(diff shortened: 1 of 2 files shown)


TwitchBronBron (changes requested):
Let's add unit tests for all of these new "default" or "user value" assignments, now that the .getOptions function is gone, each function needs to be tested with its own stuff

TwitchBronBron (approved):

PR #291, opened 2026-06-10
Allow setting options when creating the Roku-Deploy object

### LINKED ISSUE #274: Remove logLevel as a function-level option
several of our rokuDeploy functions accept `logLevel` as an option. This is silly. If you want to manipulate the logLevel, create a new instance of RokuDeploy, and set the logLevel in the constructor. (related: make sure we can pass loglevel into the rokudeploy constructor). 

Another item to consider, is the idea that you could pass a `Logger` INTO RokuDeploy instead of specifying a logLevel, and that would be the logger RokuDeploy uses instead. 
https://github.com/rokucommunity/roku-deploy/issues/150#issue-2167702191

### LINKED ISSUE #275: Specify defaults in the RokuDeploy constructor to skip passing them to each method
Support passing high-level default properties to roku-deploy so that you don't need to specify them in calls later on. (i.e. you already specified host in the constructor, so you wouldn't need to pass that when calling `.sideload()`). 

```typescript
new RokuDeploy({
    host: '1.2.3.4',
    password: 'aaaa'
    ecpPort: 8060
});
```

Open questions:
- how do we handle specific function interfaces? host is _required_ to run `.sideload()`, except when pre-populated in the constructor. so that makes things a little confusing. maybe there's a typescript trick we can do where, if you specify a field in the constructor, we return a different _type_ back that narrows the options on the functions accordingly? Might not be worth the complexity...

Christian-Holbrook:
Add constructor to `RokuDeploy` that accepts default options, so callers don't need to repeat `host`, `password`, etc. on every method call. Also adds       
  support for custom logger instances.                                                                                            
                                         
  Closes #275, closes #274                                                                                                                                     
   
  ---                                                                                                                                                          
                                                          
  ### Before                             

  ```typescript
  const rd = new RokuDeploy();

  await rd.sideload({ host: '192.168.1.100', password: 'secret', dir: './dist' });
  await rd.deleteDevChannel({ host: '192.168.1.100', password: 'secret' });
  await rd.captureScreenshot({ host: '192.168.1.100', password: 'secret', out: './screenshot.png' });
```

  ### After
  ```typescript
  const rd = new RokuDeploy({ host: '192.168.1.100', password: 'secret' });

  await rd.sideload({ dir: './dist' });
  await rd.deleteDevChannel();
  await rd.captureScreenshot({ out: './screenshot.png' });

  Custom Logger

  const rd = new RokuDeploy({ logger: myCustomLogger });
  // or just set log level
  const rd = new RokuDeploy({ logLevel: 'debug' });

### CODE REVIEW

File: src/RokuDeploy.ts [hunk 1]
```diff
@@ -29,12 +29,57 @@ export class RokuDeploy {
...
+
+        // Use custom logger if provided, otherwise use global logger
+        if (this.options.logger) {
+            this._logger = this.options.logger;
+        } else {
+            this._logger = logger;
+        }
+
+        // Set log level if provided
+        if (this.options.logLevel) {
```
Christian-Holbrook: We decided to drop logLevel as an option. We will unify with other tools that just take logger.

See example from `brighterscript`
```
/**
 * A runner class that handles
 */
export class ProgramBuilder {

    public constructor(
        options?: {
            logger?: Logger;
        }
    ) {
        this.logger = option
```

src/RokuDeploy.ts
```diff
@@ -29,12 +29,57 @@ export class RokuDeploy {
...
+     * These defaults will be merged into every method call, so you don't need to
+     * repeat common options like `host` and `password` on every call.
+     *
+     * @example
+     * ```typescript
+     * const rd = new RokuDeploy({ host: '192.168.1.100', password: 'secret' });
+     * await rd.sideload({ rootDir: './dist' }); // host and password come from defaults
+     * ```
+     */
+    constructor(options?: RokuDeployOptions) {
```
TwitchBronBron: Let's make a new interface called something like `RokuDeployConstructorOptions` that is just a few curated props that are the same name in our functions and represent the same prop across them. 

Some examples are:
- host
- password
- username
- packagePort 
- remotePort
- maybe some networking options (eventually? ) like timeouts and stuff, but those can come later...

src/RokuDeploy.spec.ts
```diff
@@ -4294,6 +4294,41 @@ describe('RokuDeploy', () => {
     }
 
     describe('defaults', () => {
+        describe('constructor defaults', () => {
```
TwitchBronBron, line 4961: We'll want tests that prove that each option is used in all the functions that reference them. probably gonna be a decent number of little extra tests. 

test like:
 - didn't construct with options, didn't provide option in call, so fail
 - provided in constructor, didn't provide in call, use constructor one
 - provided in constructor, provided in call, use call one
 - not provided in constructor, provided in call, use call one

src/RokuDeploy.ts
TwitchBronBron: Let's eliminate the concept of `.logLevel` in favor of setting that on logger itself. (make sure logger is public)

So external callers would do rokuDeploy.logger.logLevel = whatever

src/RokuDeploy.ts
```diff
@@ -29,12 +29,57 @@ export class RokuDeploy {
...
+        // Set log level if provided
+        if (this.options.logLevel) {
+            this._logger.logLevel = this.options.logLevel;
+        }
+    }
+
+    /**
+     * Get the logger instance for this RokuDeploy instance
+     */
+    private get log() {
```
TwitchBronBron: don't need this. just make `this._logger` public `this.logger`

TwitchBronBron (changes requested):

src/RokuDeploy.spec.ts
Christian-Holbrook: Do another pass to make sure _each function_ is using the correct property.

src/RokuDeploy.ts
```diff
@@ -29,12 +29,45 @@ export class RokuDeploy {
         outFile: 'roku-deploy.zip'
     };
 
+    /**
+     * Instance-level default options that are merged into every method call
+     */
+    private readonly options: RokuDeployConstructorOptions;
+
+    /**
```
Christian-Holbrook, line 68: These comments are a little verbose. Trim them down

TwitchBronBron (approved):

PR #295, opened 2026-06-25
Move load config from util to RokuDeploy

Christian-Holbrook:


Christian-Holbrook:
Rename loadOptionsJson to loadConfigFile

TwitchBronBron (approved):
Let's make `loadConfigFile` not static. after that, you can merge.

PR #300, opened 2026-06-30
Extract normalizeDeviceInfo helper from getDeviceInfo

chrisdp:
Exposes the device-info enhancement (camel-case keys plus value normalization) as a public `enhanceDeviceInfo(DeviceInfoRaw): DeviceInfo` method, so callers that already have a raw device-info object can enhance it without making another request to the device. `getDeviceInfo({ enhance: true })` now delegates to it.

Used by roku-debug and the vscode extension to reuse already-gathered device-info during debug launch instead of issuing a redundant request.

### DIFF
--- a/src/RokuDeploy.ts
+++ b/src/RokuDeploy.ts
@@ -1311,12 +1311,7 @@ export class RokuDeploy {
             } as Record<string, any>;
 
             if (options.enhance) {
-                const result = {};
-                // sanitize/normalize values to their native formats, and also convert property names to camelCase
-                for (let key in deviceInfo) {
-                    result[util.camelCase(key)] = this.normalizeDeviceInfoFieldValue(deviceInfo[key]);
-                }
-                deviceInfo = result;
+                return this.normalizeDeviceInfo(deviceInfo);
             }
             return deviceInfo;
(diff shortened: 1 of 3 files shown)


src/RokuDeploy.ts
```diff
@@ -1348,6 +1343,23 @@ export class RokuDeploy {
...
 
+    /**
+     * Enhance a raw device-info object into its normalized form. This camel-cases the property names and
+     * normalizes each value to its native format (boolean strings to booleans, number strings to numbers,
+     * decoding HtmlEntities, etc.). This is the same enhancement `getDeviceInfo` applies when called with
+     * `{ enhance: true }`, exposed separately so callers that already have a raw device-info object can
+     * enhance it without making another request to the device.
+     * @param deviceInfo the raw device-info object to enhance
+     */
+    public normalizeDeviceInfo(deviceInfo: DeviceInfoRaw): DeviceInfo {
```
TwitchBronBron: I think I want to call this `enhanceDeviceInfo` since the property on `getDeviceInfo` is `enhance`

TwitchBronBron (approved):

PR #304, opened 2026-07-02
Add option to get screenshot as a buffer

### LINKED ISSUE #172: Support getting screenshot as stream instead of file
Add support for `getScreenshot()` returning a buffer instead of a file path, for apps that would leverage the buffer directly instead of needing to load the file from disk.

Christian-Holbrook:
Adds returnBuffer option to captureScreenshot() that returns a Buffer instead of writing to disk.

  Closes #172                                                                                                        
   
  ---                                                                                                                
  API                                                      
                                                          
```typescript
  // Existing behavior - returns file path
  const path = await rokuDeploy.captureScreenshot({ host, password });


  // New behavior - returns Buffer
  const buffer = await rokuDeploy.captureScreenshot({ host, password, returnBuffer: true });
```

  ---
  ### Why Overloads?

  TypeScript overloads provide proper return type inference based on the options passed:

  ```typescript
// TypeScript knows this is Promise<Buffer>
  const buffer = await captureScreenshot({ host, password, returnBuffer: true });

  // TypeScript knows this is Promise<string>
  const path = await captureScreenshot({ host, password });
```

  Without overloads, the return type would be Promise<string | Buffer>, requiring users to narrow the type manually:

```typescript
  // Without overloads - awkward
  const result = await captureScreenshot({ host, password, returnBuffer: true });
  if (Buffer.isBuffer(result)) {
    // now we can use it as Buffer
  }
```

  This pattern follows the existing convention in the codebase (see `getDeviceInfo()` with its enhance option).

### DIFF
--- a/src/RokuDeploy.spec.ts
+++ b/src/RokuDeploy.spec.ts
@@ -3206,7 +3206,7 @@ describe('RokuDeploy', () => {
             let result = await rokuDeploy.captureScreenshot({ host: options.host, password: 'password', out: `${tempDir}/screenshot.png` });
             expect(result.buffer).to.be.instanceOf(Buffer);
             expect(result.buffer.toString()).to.equal('fake-image-data');
-            expect(result.filePath).to.equal(`${tempDir}/screenshot.png`);
+            expect(result.filePath).to.equal(path.join(tempDir, 'screenshot.png'));
             expect(fsExtra.existsSync(result.filePath)).to.be.true;
         });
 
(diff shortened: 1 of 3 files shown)


Christian-Holbrook:
Change the function to always return an object with a property called buffer.

There is an optional parameter called "out" that can be true or the path to save the screenshot. That adds the file path to the return object.

TwitchBronBron (approved):

PR #325, opened 2026-07-20
Full test coverage

### LINKED ISSUE #151: Re-enable and hit 100% coverage
As part of the v4 rewrite, we disabled code coverage to keep the PRs moving. Before finalizing the v4 release we need to reenable coverage checking and hit 100% coverage again.

Christian-Holbrook:


src/fetch.spec.ts
```diff
@@ -38,4 +39,104 @@ describe('fetch module', () => {
...
+                challenge: {}
+            });
+            expect(emptyValuesResult).to.include('realm=""');
+            expect(emptyValuesResult).to.include('nonce=""');
+        });
+    });
+
+    describe('parseDigestChallenge', () => {
+        it('parses quoted values', () => {
+            const { parseDigestChallenge } = require('./fetch');
```
TwitchBronBron: why not just import `parseDigestChallenge` at the top of the file instead? don't love seeing `require` calls.

TwitchBronBron (approved):

PR #330, opened 2026-07-23
Expose public defaults and output path resolvers

chrisdp:
With `getOptions()` gone in v4, consumers (roku-debug and the vscode extension, currently being migrated) have no way to resolve the defaults roku-deploy uses internally, so they end up hardcoding `.roku-deploy-staging` and `roku-deploy.zip` strings.

- Make `RokuDeploy.defaults` public and add `stagingDirName`, `username`, and `files` (the existing `DefaultFiles`)
- Add `getStagingDir`, `getOutputZipPath`, and `getOutputPkgPath` resolvers (`{out?, outDir?, outFile?, cwd?}`, where `out` wins; zip/pkg extension enforcement included). `stage`, `zip`, `sideload`, and `createSignedPackage` now resolve their paths through these same functions, so the library and its consumers can't drift
- Make `RekeyDeviceOptions.devId` optional: `rekeyDevice` only uses it as a post-rekey verification and `checkRequiredOptions` doesn't include it
- Replace the unused `GetOutputZipFilePathOptions` interface with the new resolver option interfaces

### DIFF
--- a/src/RokuDeploy.ts
+++ b/src/RokuDeploy.ts
@@ -30,16 +30,35 @@ import * as semver from 'semver';
 import { fetchWithDigest } from './fetch';
 import { formatTimestampForScreenshot } from './dateUtils';
 
+export const DefaultFiles = [
+    'source/**/*.*',
+    'components/**/*.*',
+    'images/**/*.*',
+    'locale/**/*',
+    'fonts/**/*',
+    'manifest',
+    '!node_modules',
+    '!**/*.{md,DS_Store,db}'
+];
+
 export class RokuDeploy {
     /**
-     * Default values for common options used across multiple functions
+     * Default values for common options used across multiple functions.
+     * Public so consumers can resolve the same defaults roku-deploy uses internally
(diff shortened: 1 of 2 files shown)


src/RokuDeploy.ts
```diff
@@ -30,16 +30,35 @@ import * as semver from 'semver';
...
         packagePort: 80,
         ecpPort: 8060,
+        username: 'rokudev',
         outDir: './out',
-        outFile: 'roku-deploy.zip'
+        outFile: 'roku-deploy.zip',
+        /**
+         * The name of the staging folder that gets created inside `outDir`
+         */
+        stagingDirName: '.roku-deploy-staging',
```
TwitchBronBron, line 60: I'd like to find a way to not expose `outFile` as literally the name `outFile`. We've aligned on `out` as the name for several of our functions, so this gets a bit too ambiguous. 

One possible solution to that is to structure the defaults like this:

```ts
defaults = {
    stage: {
        out: '.roku-deploy-staging`
    },
    zip: {
        out: 'roku-deploy.zip`
    }
}
```

or, lean more into the helper functions directly `getStagingDir`, instead of exposing default object at all.

PR #331, opened 2026-07-24
Roku Cloud Emulator support

chrisdp:
Adds Roku Cloud Emulator (RCE) device support to roku-deploy, backing the extension's RCE integration (rokucommunity/vscode-brightscript-language#861).

Stacked on #323 (unified device option).

- `RceManagementClient`: typed client for the RCE management API (devices, lifecycle, snapshots, firmware, logs, user), with `DeviceRun` matching the full device-history response shape. Every call takes an options object accepting an optional per-call `token` that overrides the constructor token (one client can serve multi-account callers), and `getInstanceUrl` resolves a device config addressed by instanceUrl, id, or esn to its live instance URL. Device ids are the management api's numeric ids (`id: number` on the config, with numeric strings from untyped json configs coerced and anything else rejected client-side). Esn resolution searches the whole device inventory (`items=0`, the api's no-limit value) instead of the first page, and the paginated list endpoints (devices, firmware versions, snapshots) expose `items`/`page`. Transport hygiene matches `request.ts`: the per-request timeout bounds both connection and first response byte, and each request closes its socket instead of pooling it, so a CLI process exits when its work is done
- Device-level RCE additions on `RokuDeploy` (replacing the earlier `RceDevice` class): a constructor-level default `rceToken` for RCE device configs that carry none, memoized instance-url resolution per device+token, `resolveDns(device)` (resolves a local device config's host to its ip; other device kinds pass through unchanged), `sendKeySequence` (presses remote keys in order through the same transport as `keyPress`, so key macros work on local and cloud devices alike), and `sendDeveloperSettingsCombo` (the RCE instance-api developer-settings action)
- `createRokuDeploySocket({ device, port })`: a near-transparent `net.Socket` replacement for a device's tcp ports (the BrightScript console, the SceneGraph debug server, the debug protocol's control and io ports); a local device connects plain tcp, and an RCE device reaches the same port over the instance api's `/api/v0/ports/<port>` websocket routes (Instance API v0.3.0; binary frames both ways, byte parity with a raw tcp socket). Lifecycle semantics match `net.Socket`: writes issued while connecting are buffered and flushed on open, `end()` runs the websocket close handshake, a remote close delivers `'end'` before `'close'`, a full readable buffer pauses the websocket until the consumer drains it, `'close'` carries `hadError`, an explicit `wss:` instanceUrl is never downgraded, and a second `connect()` throws instead of orphaning the live websocket
- `RceVideoSignalingClient` (the Janus signaling client the extension's video streams ride) hardened on the abnormal paths: a failed or cancelled `connect()` tears the session down instead of leaking the keepalive timer and socket, an unexpected socket close rejects the requests still in flight instead of hanging their callers, and a second `connect()` on an active session rejects (reconnect after `stop()` works)
- exports `DeviceConfig`, `RceManagementClient`, `RceVideoSignalingClient`, and `createRokuDeploySocket` from the package index
- Remote key input (`keyPress`/`keyUp`/`keyDown`/`sendText`) prefers the RCE instance-api key route (`/api/v0/ecp1/<action>/<key>`), which keeps working while the device is in limited ECP mode (the raw `ecp1` proxy 403s key presses there) and falls back to the raw proxy otherwise (live-verified in limited mode). Keys are exposed as a `RemoteKey` enum in canonical device casing, with a `RemoteKeyText` type enforcing the canonical names on the public key methods; this replaces the old lowercase `RokuKey` type
- `ecp(device, route)`: the single public ECP transport, addressing local devices on the HTTP ECP port and RCE devices through the instance's `ecp1` proxy, always returning the XML body parsed to JSON. Dedicated wrappers (`getDeviceInfo`, key presses, `launchApp`, `exitApp`, `queryApps`, `queryActiveApp`, `queryRegistry`, `queryAppState`, `queryRendezvous`, `setRendezvousTracking`) all funnel through it with their `appId` uri-encoded, and device-reported failure text (not keyed, limited-mode refusals) rides in the typed errors. ECP requests default to a 10s `RokuDeploy.defaults.ecpTimeout` (overridable per call), separate from the 150s installer timeout. Live-verified against a cloud device, including in limited ECP mode


src/RceManagementClient.ts
```diff
@@ -0,0 +1,363 @@
...
+                if (statusCode < 200 || statusCode >= 300) {
+                    reject(new Error(`RCE management ${method.toUpperCase()} ${path} failed (status ${statusCode})`));
+                    return;
+                }
+                resolve(response.body as TResponse);
+            });
+        });
+    }
+
+    private buildQueryString(query?: Record<string, string | number | undefined>): string {
```
chrisdp, line 227: Likely could be a util. Should check and see if we have one already and if not move it there. Also should verify if still needed after the change to using auth headers.

src/RceManagementClient.ts
```diff
@@ -0,0 +1,363 @@
...
+    janus_token?: string | null;
+    janus_websocket_url?: string | null;
+    janus_ice_servers?: IceServer[] | null;
+    instance_api_url?: string | null;
+    instance_uuid: string;
+    firmware_version_id: string;
+    max_runtime: number;
+}
+
+export interface DeviceOut {
```
chrisdp: Something about these interface names bother me. Part of it might be using out rather then response or something like that. 

I want to revisit these with @TwitchBronBron

src/RceManagementClient.ts
```diff
@@ -0,0 +1,363 @@
...
+    username?: string | null;
+    credential?: string | null;
+}
+
+export interface DeviceInstanceInfo {
+    id: number;
+    creator_id: string;
+    created_at: string;
+    started_at?: string | null;
+    snapshot_id: number;
```
chrisdp, line 384: @TwitchBronBron in general what are your thoughts about the property names? Was almost thinking about making a small utility that runs every prop through a rename to camel case. Maybe overkill be let me know.

src/RceDevice.ts
```diff
@@ -0,0 +1,153 @@
...
+     * 'Select', ...); the route rejects a wrong-case key with a 422. Each press waits for the
+     * previous press's response plus a small delay (keyDelayMs) so on-screen navigation keeps up;
+     * the first failed press throws with the failing key and step.
+     *
+     * This deliberately does NOT go through the raw `/ecp1` proxy RokuDeploy's `ecp()` transport
+     * uses: that proxy forwards unauthenticated ECP, which a device in limited ECP mode rejects with
+     * an empty 403. The instance-api key route carries the instance's own authentication, so key
+     * input works regardless of the device's ECP mode - which is what makes key macros like the
+     * disable-limited-ECP sequence possible in the first place.
+     */
```
chrisdp: Some comments are a bit verbose and in particular some doc blocks are digging deeper into the underlying implementation then the consumer needs to care about. Want to do a pass on these.

src/RokuDeploy.ts
```diff
@@ -20,28 +20,54 @@ import {
...
+     * instead of hardcoding them (see also `getStagingDir`, `getOutputZipPath` and `getOutputPkgPath`).
      */
-    private static readonly defaults = {
+    public static readonly defaults = {
         timeout: 150000,
+        /**
+         * The default timeout for ECP requests, in milliseconds. ECP commands are small and fast,
+         * so this is much lower than `timeout` (which must accommodate large installer uploads).
+         */
+        ecpTimeout: 10000,
```
chrisdp, line 48: @TwitchBronBron added a separate timeout for ECP as I think the universal default we had is a little nuts for ECP calls. Let me know if you agree that they can be shorter then sideloading for example and if your happy with the 10s I set.

src/RokuDeploy.ts
```diff
@@ -1794,24 +2357,66 @@ export interface ValidateDeveloperPasswordOptions {
...
+    Up = 'Up',
+    VolumeDown = 'VolumeDown',
+    VolumeMute = 'VolumeMute',
+    VolumeUp = 'VolumeUp'
+}
+
+export type RemoteKeyText = keyof typeof RemoteKey;
 
-export interface SendKeyEventOptions extends BaseEcpOptions {
+interface SendKeyEventOptions extends BaseEcpOptions {
```
chrisdp: Should we have removed the export here?

src/RokuDeploy.ts
```diff
@@ -92,6 +99,14 @@ export class RokuDeploy {
      */
     public readonly logger: typeof logger;
 
+    /**
+     * One resolved instance url per unique RCE device config, cached for the lifetime of this
+     * `RokuDeploy` instance, so a multi-request flow (for example sideload's closeChannel ->
+     * deleteDevChannel -> plugin_install) avoids re-resolving the instance url through the
+     * management api on every request. Failed resolutions are evicted so the next call retries.
+     */
+    private readonly rceInstanceUrlsByCacheKey = new Map<string, Promise<string>>();
```
TwitchBronBron, line 109: In a followup PR, let's add "try cache, it failed, get fresh value" flow for this instead of "fail this call but the next one will succeed"

src/RokuDeploy.ts
```diff
@@ -298,6 +314,216 @@ export class RokuDeploy {
...
+    /**
+     * Resolve the request base (base url + headers that must be sent with every request) for reaching a
+     * device's installer. Local devices hit the classic port-80 dev installer directly; RCE devices are
+     * reached through the instance api's `/sideload` proxy, which forwards to the emulated device's installer.
+     *
+     * The RCE instance api sits behind a service mesh that authenticates the bearer token via the
+     * `X-Authorization` header (rather than the standard `Authorization` header, which is reserved for
+     * the installer's own HTTP Digest challenge) - two independent auth layers that can't share the
+     * `Authorization` header.
+     */
```
TwitchBronBron, line 327: Please, in a followup PR, condense these comments. We're human. ;)

src/RokuDeploy.ts
```diff
@@ -298,6 +314,216 @@ export class RokuDeploy {
...
+    }
+
+    /**
+     * Resolve a device config's host through DNS, returning a new config with the host replaced by
+     * its ip address (some Rokus reject ECP requests addressed by hostname or mDNS name, so callers
+     * use this to pin a config to an ip up front). Only local devices are addressed by host; any
+     * other device config (like a Roku Cloud Emulator device) is returned unchanged. A failed
+     * lookup throws so the caller decides how to handle an unreachable host.
+     */
+    public async resolveDns<T extends DeviceConfig>(device: T): Promise<T> {
```
TwitchBronBron: Let's find a better name for this (can be in a followup PR).

src/RokuDeploy.ts
```diff
@@ -298,6 +314,216 @@ export class RokuDeploy {
...
+     * Local devices are addressed as `http://<host>:<ecpPort>/<route>`; Cloud Emulator devices go
+     * through their instance's `/ecp1/<route>` proxy (authenticated with the config's api token),
+     * so callers never branch on device kind.
+     * @param device the device to send the request to
+     * @param route the ECP route without a leading slash (for example `query/device-info`, `keypress/Home`)
+     * @param options `method` defaults to 'GET' (queries); commands like keypress and launch are POSTs.
+     *                `verify` defaults to false so raw ECP status bodies (a 202 `FAILED` registry or
+     *                chanperf response, for example) come back to the caller instead of throwing.
+     */
+    public async ecp(device: DeviceOption, route: string, options?: EcpOptions): Promise<EcpResult> {
```
TwitchBronBron: Find a better name for this. something like `.sendEcpRequest`?

src/RokuDeploy.ts
```diff
@@ -298,6 +314,216 @@ export class RokuDeploy {
...
+        const requestOptions: RequestOptions = { url: `${baseUrl}/${route}`, timeout: timeout, headers: headers };
+        const verify = options.verify ?? false;
+        const response = options.method === 'POST'
+            ? await this.doPostRequest(requestOptions, verify)
+            : await this.doGetRequest(requestOptions, verify);
+
+        return {
+            status: response?.response?.statusCode,
+            body: response?.body,
+            json: await this.parseEcpXml(response?.body, response)
```
TwitchBronBron, line 462: might be better to just skip sending JSON for this. but we can address that before v4 lands, so it's fine here for now

src/RokuDeploy.ts
```diff
@@ -298,6 +314,216 @@ export class RokuDeploy {
...
+            throw new Error('An rceToken is required to resolve an RCE device by id or esn');
+        }
+        return this.createRceManagementClient(rceToken).getInstanceUrl({ device: deviceConfig });
+    }
+
+    /**
+     * Create the management client used to resolve an id- or esn-addressed RCE device to its running
+     * instance URL. Split out so tests can supply a fake.
+     */
+    protected createRceManagementClient(rceToken: string): RceManagementClient {
```
TwitchBronBron: should be `private`

src/RokuDeploy.ts
```diff
@@ -337,27 +563,116 @@ export class RokuDeploy {
...
+     */
+    public async sendKeySequence(options: SendKeySequenceOptions): Promise<void> {
+        options = { ...this.options, ...options } as SendKeySequenceOptions;
+        this.checkRequiredOptions(options, ['device', 'keys']);
+        const keyDelayMs = options.keyDelayMs ?? 250;
+        for (let stepIndex = 0; stepIndex < options.keys.length; stepIndex++) {
+            if (stepIndex > 0 && keyDelayMs > 0) {
+                await new Promise((resolve) => {
+                    setTimeout(resolve, keyDelayMs);
+                });
```
TwitchBronBron: use `util.sleep()`

src/RokuDeploy.ts
```diff
@@ -337,27 +563,116 @@ export class RokuDeploy {
...
+    }
+
+    /**
+     * Enter the developer-settings key combo on a Roku Cloud Emulator device through its instance
+     * api (the same combo a physical remote's key sequence would send). The device then shows the
+     * on-screen developer setup wizard for the user to complete; this call only triggers that
+     * screen, it does not finish the setup itself. Local devices are not supported - the combo
+     * endpoint only exists on the RCE instance api.
+     */
+    public async sendDeveloperSettingsCombo(options: SendDeveloperSettingsComboOptions): Promise<void> {
```
TwitchBronBron, line 713: Not loving this name. maybe `enableDeveloperMode` or `invokeDeveloperMode`? More focused on "what" it's doing instead of "how".

or, we could instead move more to like a `.sendDeveloperAction` which takes a property.

src/RokuDeploy.ts
```diff
@@ -1324,7 +1725,8 @@ export class RokuDeploy {
         if (entry.esn) {
             return { esn: entry.esn, rceToken: entry.rceToken };
         }
-        if (entry.id) {
+        //an explicit !== undefined check (unlike the truthy string checks above): 0 is a valid number
+        if (entry.id !== undefined) {
             return { id: entry.id, rceToken: entry.rceToken };
         }
         if (entry.instanceUrl) {
```
TwitchBronBron: Let's use the helpers from DeviceConfig, and ensure those helpers do any logic from this function that are necessary. (this unifies the experience from public and private usage)

src/RokuDeploySocket.ts
```diff
@@ -0,0 +1,524 @@
...
+ * A `net.Socket` wired up to connect to a local device's plain-tcp telnet console using the host and
+ * port resolved by `createRokuDeploySocket()`. Every behavior other than `connect()` is real
+ * `net.Socket` behavior inherited unchanged; that is the entire point of extending it directly
+ * instead of wrapping it in another layer.
+ */
+export class LocalSocket extends net.Socket {
+    constructor(
+        private readonly host: string,
+        private readonly port: number
+    ) {
```
TwitchBronBron: I think I might want to have a single options constructor arg instead of ordered params. Same with the RceSocket

TwitchBronBron (approved):

src/RceManagementClient.ts
```diff
@@ -0,0 +1,518 @@
...
+    /**
+     * Runtime of the device instance, in seconds.
+     */
+    runtime?: number;
+    firmware_version_id?: string | null;
+    /**
+     * The maximum runtime allowed for the device instance, in seconds.
+     */
+    max_runtime?: number;
+    [key: string]: unknown;
```
TwitchBronBron, line 464: Why is the indexer here, but only for this response?

src/RceManagementClient.ts
chrisdp: Talked and we will align on camel everywhere and convert snake on the way out and camel on the way in for post bodies

src/RceManagementClient.ts
chrisdp: Lets be strict

src/RokuDeploy.ts
```diff
@@ -429,12 +430,31 @@ export class RokuDeploy {
...
-        const requestOptions: RequestOptions = { url: `${baseUrl}/${route}`, timeout: timeout, headers: headers };
         const verify = options.verify ?? false;
-        const response = options.method === 'POST'
-            ? await this.doPostRequest(requestOptions, verify)
-            : await this.doGetRequest(requestOptions, verify);
+        //with verify off, an instance-gone 404 comes back as a normal response rather than a throw;
+        //it is surfaced to the retry wrapper as an error, and held here so the caller still receives
+        //the raw response when the retry does not apply
+        let instanceGoneResponse: HttpResponse | undefined;
+        const response = await this.withRceInstanceUrlRetry(deviceConfig, async () => {
```
TwitchBronBron, line 438: let's add a comment explaining that all requeests are tried once, then if rce cache bust happens, try one more time.

src/DeviceConfig.ts
```diff
@@ -56,36 +57,46 @@ export type DeviceOption = string | DeviceConfig;
 // === Type Guards ===
 
 /**
- * Check if a device config is for a local network device
+ * Any object that may carry device identifier keys (a DeviceConfig, a registry entry, etc.).
```
TwitchBronBron:
```suggestion
 * Any object that may carry device identifier keys (a DeviceConfig, a registry entry, etc.)
```

chrisdp:
@copilot please fix these type/import errors in the src/DeviceConfig.spec.ts file. 

2026-08-07T00:15:10.2499485Z ##[error]src/DeviceConfig.spec.ts(2,50): error TS2305: Module '"./DeviceConfig"' has no exported member 'isRceByEsn'.
2026-08-07T00:15:10.2509051Z ##[error]src/DeviceConfig.spec.ts(2,62): error TS2305: Module '"./DeviceConfig"' has no exported member 'isRceById'.
2026-08-07T00:15:10.2510876Z ##[error]src/DeviceConfig.spec.ts(2,73): error TS2305: Module '"./DeviceConfig"' has no exported member 'isRceByUrl'.
2026-08-07T00:15:10.2512523Z ##[error]src/RokuDeploy.spec.ts(6201,39): error TS2322: Type 'string' is not assignable to type 'number'.
2026-08-07T00:15:10.2758452Z ##[error]Process completed with exit code 2.

PR #354, opened 2026-08-06
Chore: Hard-reboot device via smart switch for device tests

TwitchBronBron:
## Summary
- Adds `src/smartSwitchManagement.spec.ts`, which power-cycles the Roku via a local smart plug (protocol 3.4, LAN-only, no cloud calls at test time) for a true hard reset when the device might be wedged.
- Used only by the suite-level `before`/`after` health checks, and only when `SMART_SWITCH_DEVICE_ID`/`SMART_SWITCH_LOCAL_KEY`/`SMART_SWITCH_IP` are configured (see updated `.env.example`). Falls back to the existing software `rebootDevice` ECP call if unset, so this is fully opt-in and doesn't require every contributor to own the same hardware.
- Consolidates the old separate "mid-suite reboot" check into the real `rebootDevice` test, relocated to the suite's actual midpoint (previously it ran twice for no added coverage).

## Test plan
- [x] `npm run test:device` run against a real device with `SMART_SWITCH_*` configured — confirmed the plug physically power-cycles the Roku before and after the suite.
- [x] `npx tsc --noEmit` / `npm run lint` clean.


src/device.spec.ts
```diff
@@ -542,6 +530,29 @@ describe('device', function device() {
         });
     });
 
+    //placed at roughly the suite's midpoint (rather than at the end) so a device that can't keep up
```
TwitchBronBron:
```suggestion
    //placed at roughly the suite's midpoint (rather than at the end) so a device that isn't able to keep up
```

PR #30, opened 2019-11-14
Add packagePort and remotePort options

TwitchBronBron:
Fixes #29.

PR #360, opened 2026-08-19
Reduce Public Api Surface

### LINKED ISSUE #335: Reduce public API surface and implement API Extractor
Similar to https://github.com/rokucommunity/logger/pull/30, we should strip out any "internal" features we don't want exposed as public.

Christian-Holbrook:
Implement API Extractor & reduce public API surface

Closes #335. Mirrors logger PR #30: adds API Extractor with a committed API report (api/roku-deploy.api.md) and a trimmed dist/index.d.ts rollup, then aggressively trims the public surface (190 → 131 exports). A guard spec (src/apiExtractor.spec.ts, run in both build and test) fails on untagged exports, leaked internal types, or report drift.

BREAKING CHANGES

Removed from the public type surface (still shipped as JS, just no longer typed/exported):

- util — Util, defer, Deferred (standardizePath / standardizePathPosix stay public)
- RCE management — RceManagementClient and its option/model types
- RCE video signaling — RceVideoSignalingClient and its types
- Sockets — createRokuDeploySocket, RokuDeploySocket, LocalSocket, RceSocket and option types
- RokuDeploy — checkRequiredOptions is now @internal

Notable changes for review

- typescript ^5.4.3 → ~5.4.3 — keeps the build on 5.4.x; API Extractor uses its own bundled 5.9.3 (5.9 is incompatible with the pinned @types/node@18).
- skipLibCheck: true — so API Extractor's 5.9.3 analysis ignores @types/node typings.
- files allowlist — ships only dist/**/*.js, .js.map, and the rolled-up dist/index.d.ts (publish-behavior change).


src/util.ts
TwitchBronBron, line 1: Let's do a pass on all the functions in `class Util`, and lock most of them down as `@internal` (if we can).

src/RokuDeploy.ts
```diff
@@ -1486,6 +1489,7 @@ export class RokuDeploy {
         return result;
     }
 
+    /** @internal */
```
TwitchBronBron: Can't this just be changed to `private` method? I'm okay with unit tests using square access to bypass typescript private errors.

TwitchBronBron (changes requested):

src/util.ts
Christian-Holbrook: I have done a pass on this. Fixed some conflicts. This should be ready for a review again.

TwitchBronBron (changes requested):
I haven't gotten a chance to review this yet, but I'd like to see every single `export`ed item have a @public, @internal, or @private annotation on it. this pr feels maybe al ittle lite in that regard? A good example are the commands. Those classes have no jsdoc comment at all.

Can we find an eslint rule that enforces that every exported thing must have a public/protected/private marker? That might help surface exactly what's being exposed.

src/RokuDeploy.ts
TwitchBronBron, line 682: Another example, these methods are public, but we should have a @public on them to clarify that they are indeed intended to be exposed to the public api.

TwitchBronBron (approved):

Issue #353, opened 2026-08-06
Consider if ECP should be able to have its own timeout value


chrisdp:
@TwitchBronBron added a separate timeout for ECP as I think the universal default we had is a little nuts for ECP calls. Let me know if you agree that they can be shorter then sideloading for example and if your happy with the 10s I set.

_Originally posted by @chrisdp in https://github.com/rokucommunity/roku-deploy/pull/331#discussion_r3682628691_

PR #370, opened 2026-08-26
feat: layered per-class timeouts (#353)

Christian-Holbrook:
Fixes #353

Adds a `timeouts` option (`{ default, ecp, rceManagement }`) so each traffic class gets its own timeout. Resolution, most specific first: per-call `timeout` → `timeouts.<class>` → built-in class default (ecp 10s, rceManagement 30s) → `timeouts.default` / legacy `timeout` → 150s. Class defaults outrank the root on purpose — raising the root for slow uploads shouldn't slow fail-fast ECP traffic.

Also fixes the constructor `timeout` bleeding into ECP calls (directly, and via sideload/rebootDevice/checkForUpdate's nested ECP requests).

### DIFF
--- a/src/RokuDeploy.ts
+++ b/src/RokuDeploy.ts
@@ -63,17 +63,49 @@ export class RokuDeploy {
      * Default values for common options used across multiple functions
      */
     private static readonly defaults = {
-        timeout: 150000,
         /**
-         * The default timeout for ECP requests, in milliseconds. ECP commands are small and fast,
-         * so this is much lower than `timeout` (which must accommodate large installer uploads).
+         * Built-in timeouts (in milliseconds) per traffic class (see `resolveTimeout`)
          */
-        ecpTimeout: 10000,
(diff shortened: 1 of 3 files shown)


src/RokuDeploy.ts
```diff
@@ -63,18 +63,50 @@ export class RokuDeploy {
...
+        }
+        return this.options.timeouts?.[group] ?? RokuDeploy.defaults.timeouts[group];
+    }
+
+    /**
+     * Merge constructor options into a per-call options object. The constructor `timeout` is
+     * excluded: only a caller-supplied timeout travels with the options, and the transports resolve
+     * the rest per traffic class (see `resolveTimeout`).
+     */
+    private mergeOptions<T extends { timeout?: number }>(options: T): T {
```
TwitchBronBron: Don't love seeing a centralized "mergeOptions" thing here.

PR #371, opened 2026-08-26
Refine comments

### LINKED ISSUE #350: Refine comments
Some comments are a bit verbose and in particular some doc blocks are digging deeper into the underlying implementation then the consumer needs to care about. Want to do a pass on these.

_Originally posted by @chrisdp in https://github.com/rokucommunity/roku-deploy/pull/331#discussion_r3682612493_

Christian-Holbrook:
Closes #350

Comment-only pass over `src/`: trimmed verbose docblocks and needle-migration essays down to 1–3 lines (keeping the behavioral gotchas), deleted comments that just restated the adjacent code, and fixed a stale copy-pasted docblock on `standardizePathPosix` (now says forward slashes, not OS separators).

### DIFF
--- a/src/Errors.ts
+++ b/src/Errors.ts
@@ -22,7 +22,7 @@ export enum RokuDeployErrorCode {
 }
 
 /**
- * HTTP request details, independent of the underlying HTTP library
+ * HTTP request details
  */
 export interface HttpRequestDetails {
     url?: string;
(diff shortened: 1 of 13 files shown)


TwitchBronBron (changes requested):
Do you mind giving this another pass? There are several places where I feel like the comments have been _too_ condensed now, and we're missing some useful information. Let me know if you need me to call those spots out, but it might be worth another pass.

TwitchBronBron (approved):

PR #372, opened 2026-08-26
Add rce startDevice/stopDevice CLI commands

### LINKED ISSUE #361: Add cli commands for RCE device starting and stopping
Now that we have RCE support in the roku-deploy nodejs apis, we also should expose some functionality in the CLI. The most important commands currently are:
- start a device
- stop a device

We need these asap, but other commands can follow shortly after, since they're additive.

Christian-Holbrook:
Adds RCE device lifecycle commands to the CLI.

## Commands

- `roku-deploy rce start` — boot a device from a snapshot
- `roku-deploy rce stop` — shut down a running device

Both accept `--deviceId` or `--esn`, take the token from `--token` (falling back to `rceToken` in rokudeploy.json), and support `--wait [--timeout 600]` to poll until the device reaches `running`/`shutdown` — useful for CI.

## Defaults (mirroring the vscode extension)

- Snapshot: `--snapshot <name>` (`live` selects the live snapshot) or `--snapshotId <id>` → defaults to the live snapshot → error if none exists. The device's last-loaded snapshot is deliberately NOT a fallback: stopping a device saves its state to `live`, so booting the last-loaded snapshot would silently revert that state.
- Firmware: explicit flag → snapshot's firmware → device's firmware → first available for the device type
- `maxRuntime`: 3600 seconds

Commands are namespaced under a `rce` group so the follow-up commands (list, snapshots, logs) slot in cleanly. They call `RceManagementClient` directly rather than adding new `RokuDeploy` public methods (keeping with #335).

**Open question:** should the token also be readable from an env var (e.g. `ROKU_RCE_TOKEN`) for CI?

Closes #361


src/cli.ts
```diff
@@ -164,6 +166,38 @@ void yargs
         return new GetDevIdCommand().run(args);
     })
 
+    .command('rce', 'Manage Roku Cloud Emulator (RCE) devices', (rceBuilder) => {
+        return rceBuilder
+            .command('startDevice', 'Boot an RCE device from a snapshot', (builder) => {
```
TwitchBronBron: maybe we just call these `start` and `stop` instead?

src/commands/RceStartDeviceCommand.ts
```diff
@@ -0,0 +1,53 @@
...
+        const { options, client } = buildRceCommandContext(args);
+        const device = await resolveRceDevice(client, options);
+
+        //snapshot resolution: explicit flag, then the device's live snapshot (its current disk
+        //state), then its last snapshot
+        let snapshotId: number | undefined = options.snapshotId;
+        let snapshots: Snapshot[] | undefined;
+        if (!snapshotId) {
+            snapshots = await client.listSnapshots({ deviceId: device.id, items: 0 });
+            snapshotId = snapshots.find(x => x.live)?.id ?? device.last_snapshot_id ?? undefined;
```
TwitchBronBron: I don't think this is the correct approach. if you loaded snapshot "alpha", that will be marked as last_snapshot_id. however, whenever you stopped the device, it updates your current state to "live". so by loading the last snapshot, you will revert whatever's happening on live. 

let's default to the `live` snapshot. or...if it feels safer, we can just require a snapshot instead.

src/cli.ts
```diff
@@ -164,6 +166,38 @@ void yargs
...
+                return builder
+                    .option('token', { type: 'string', description: 'The RCE bearer token. Falls back to "rceToken" in rokudeploy.json', demandOption: false })
+                    .option('deviceId', { type: 'number', description: 'The numeric management-api id of the RCE device', demandOption: false })
+                    .option('esn', { type: 'string', description: 'The serial number (ESN) of the RCE device (instead of a deviceId)', demandOption: false })
+                    .option('snapshotId', { type: 'number', description: 'The snapshot to boot from. Defaults to the device\'s live snapshot, then its last snapshot', demandOption: false })
```
TwitchBronBron: i don't think most developers will know the id of the snapshots they want to boot from. i don't mind having a snapshotId argument, but we likely should support `snapshot?: string` here instead so they can do things like `roku-deploy rce start --snapshot live`

src/cli.ts
```diff
@@ -164,6 +166,38 @@ void yargs
...
+            }, (args: any) => {
+                return new RceStartDeviceCommand().run(args);
+            })
+            .command('stopDevice', 'Shut down a running RCE device', (builder) => {
+                return builder
+                    .option('token', { type: 'string', description: 'The RCE bearer token. Falls back to "rceToken" in rokudeploy.json', demandOption: false })
+                    .option('deviceId', { type: 'number', description: 'The numeric management-api id of the RCE device', demandOption: false })
+                    .option('esn', { type: 'string', description: 'The serial number (ESN) of the RCE device (instead of a deviceId)', demandOption: false })
+                    .option('wait', { type: 'boolean', description: 'Wait for the device to reach the \'shutdown\' status before exiting', demandOption: false })
+                    .option('waitTimeout', { type: 'number', description: 'How long --wait polls before giving up, in seconds', demandOption: false, default: 600 })
```
TwitchBronBron: maybe we just call this `timeout`? not sure what other reason we'd be waiting...

TwitchBronBron (changes requested):

src/commands/RceStartDeviceCommand.ts
Christian-Holbrook: This and the other 3 comments should be addressed

TwitchBronBron (approved):

Issue #351, opened 2026-08-06
Consider not returning json from the sendEcpRequest function


chrisdp:
might be better to just skip sending JSON for this. but we can address that before v4 lands, so it's fine here for now

_Originally posted by @TwitchBronBron in https://github.com/rokucommunity/roku-deploy/pull/331#discussion_r3723497159_

Christian-Holbrook:
Completed by #367 — `sendEcpRequest` now returns `{ status, body }` only; the typed wrappers parse XML internally.

PR #373, opened 2026-08-26
Include response headers in EcpResult

chrisdp:
Since #351, `sendEcpRequest` returns the raw body and callers parse it themselves - but they had no way to know what they were parsing. `EcpResult` now carries the response headers (lowercased names, an empty object when the transport produced no response) so callers can pick a parser from the content-type.

Motivating consumer: roku-test-automation's ECP migration onto `sendEcpRequest` dispatches on `content-type` to decide between XML and JSON parsing, matching the behavior it previously inherited from needle's auto-parsing.

### DIFF
--- a/src/RokuDeploy.spec.ts
+++ b/src/RokuDeploy.spec.ts
@@ -976,7 +976,7 @@ describe('RokuDeploy', () => {
 
             const result = await rd.sendEcpRequest({ id: 123 }, 'query/device-info');
 
-            expect(result).to.eql({ status: undefined, body: undefined });
+            expect(result).to.eql({ status: undefined, headers: {}, body: undefined });
         });
 
         it('retries against a refreshed instance url when the mesh reports the cached instance gone', async () => {
--- a/src/RokuDeploy.spec.ts
+++ b/src/RokuDeploy.spec.ts
@@ -825,6 +825,20 @@ describe('RokuDeploy', () => {
             expect(stub.getCall(0).args[0].url).to.equal('http://1.1.1.1:8060/query/device-info');
             expect(result.status).to.equal(200);
             expect(result.body).to.equal('<device-info><model-name>Roku</model-name></device-info>');
+            expect(result.headers).to.eql({});
+        });
+
+        it('returns the response headers so callers can pick a parser from the content-type', async () => {
+            sinon.stub(rokuDeploy as any, 'doGetRequest').callsFake(() => {
+                return Promise.resolve({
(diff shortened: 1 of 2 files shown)


src/RokuDeploy.spec.ts
```diff
@@ -825,6 +825,20 @@ describe('RokuDeploy', () => {
             expect(stub.getCall(0).args[0].url).to.equal('http://1.1.1.1:8060/query/device-info');
             expect(result.status).to.equal(200);
             expect(result.body).to.equal('<device-info><model-name>Roku</model-name></device-info>');
+            expect(result.headers).to.eql({});
+        });
+
+        it('returns the response headers so callers can pick a parser from the content-type', async () => {
+            sinon.stub(rokuDeploy as any, 'doGetRequest').callsFake(() => {
+                return Promise.resolve({
+                    response: { statusCode: 200, headers: { 'content-type': 'text/xml; charset="utf-8"' } },
```
TwitchBronBron, line 834: statusCode isn't the right value here for the stub.

src/RokuDeploy.ts
```diff
@@ -2797,6 +2798,8 @@ export interface EcpResult {
     status: number | undefined;
     /** The raw response body (usually XML, empty for command routes like keypress) */
     body: string;
+    /** The response headers (lowercased names), so callers can pick a parser from the content-type. Empty when the transport produced no response */
```
TwitchBronBron: This says "lower-cased names", but i don't see anything enforcing/applying that. Does needle publicly document that their headers are lower cased?

chrisdp: Will check. good catch.

src/RokuDeploy.spec.ts
chrisdp: whats the issue here?

TwitchBronBron: Maybe nothing? I thought the name of the response field was `status`, but maybe this is mocking something different?

chrisdp: yes, it's the mock for the network call. not our ECP response shape.

src/RokuDeploy.ts
chrisdp:
```suggestion
    /** 
     * The response headers (lowercased names), so callers can pick a parser from the content-type. Empty when the transport produced no response 
     * @see {@link https://nodejs.org/api/http.html#messageheaders | message.headers docs}
     */
```

TwitchBronBron (approved):

PR #375, opened 2026-08-26
Encode remote keys when building ECP URLs

### LINKED ISSUE #374: Move ECP key URI-encoding into keyPress/keyUp/keyDown
## Current behavior

`keyPress`, `keyDown`, and `keyUp` insert the key into the URL raw. The private `sendKeyEvent` funnel does no encoding on either route (the LAN `sendEcpRequest` path or the RCE `/api/v0/ecp1` path, where `toCanonicalRemoteKey` only normalizes casing). A `Lit_` key containing a space, `&`, or `#` produces a corrupt request.

Meanwhile `sendText` does its own `lit_${encodeURIComponent(char)}`, so the library already owns encoding on the text path but not on the key path.

Consumers currently work around this by pre-encoding before calling in:
- vscode-brightscript-language builds `'Lit_' + encodeURIComponent(key)` before `rokuDeploy.keyPress` ([BrightScriptCommands.ts#L865](https://github.com/rokucommunity/vscode-brightscript-language/blob/master/src/BrightScriptCommands.ts))
- roku-test-automation is about to do the same as part of its migration onto the v4 keypress functions

Two callers each carrying their own `encodeURIComponent` in front of the same function is the setup for a double-encode bug the moment anyone touches either side.

## Proposal

- Encode the key at URL-build time inside `sendKeyEvent`: `${action}/${encodeURIComponent(key)}` on the LAN route, and on the RCE route encode after `toCanonicalRemoteKey` runs.
- Remove the `encodeURIComponent` from `sendText` so it does not double-encode through the shared funnel.
- Widen the public `keyPress`/`keyUp`/`keyDown` key param from `RemoteKeyText` to `RemoteKeyText | string` (the private funnel already accepts `string`), so literal `Lit_...` keys do not require casts.

The named `RemoteKey` values are all URL-safe, so wire traffic for normal keys is byte-identical. Since v4 is still alpha this is a good window for the change.

## Coordination

Callers that pre-encode must drop it when this lands, otherwise `&` goes out as `Lit_%2526` and the device types `%26` literally:
- vscode-brightscript-language `BrightScriptCommands.ts` (one-line change)
- roku-test-automation (will keep pre-encoding until this lands)

## Related

`sendText` iterates with `text.split('')`, which splits astral characters (emoji) into lone surrogates and makes `encodeURIComponent` throw `URIError: URI malformed`. Iterating with `for...of` walks code points and fixes it; worth doing in the same pass since it is the same function.

chrisdp:
`sendKeyEvent` now URI-encodes the key when it builds the ECP URL, on both routes:

- RCE instance-api route: canonicalizes via `toCanonicalRemoteKey`, then encodes the result into the URL
- direct ECP route: encodes the key into the `${action}/` route

`sendEcpRequest` itself is unchanged.

Previously the key went into the URL raw, so a `Lit_` key containing a space, `&`, or `#` produced a corrupt request, and consumers had to pre-encode as a workaround. That workaround now needs to be removed wherever it exists, since encoding twice will double-encode the key. Consumer updates for vscode-brightscript-language and roku-test-automation are prepared and will land alongside.

`sendText` dropped its own per-character `encodeURIComponent` (it would now double-encode through `sendKeyEvent`) and switched from `text.split('')` to `for...of`, fixing a bug where astral characters (emoji) were split into lone surrogates and `encodeURIComponent` threw `URIError: URI malformed`.

`KeyUpOptions`, `KeyDownOptions`, and `KeyPressOptions.key` are widened from `RemoteKeyText` to `RemoteKeyText | string`, matching `SendKeyEventOptions.key`, so literal `Lit_...` keys no longer need a cast.

Closes #374

### DIFF
--- a/src/RokuDeploy.spec.ts
+++ b/src/RokuDeploy.spec.ts
@@ -2242,6 +2242,19 @@ describe('RokuDeploy', () => {
             expect(stub.getCall(0).args[0].url).to.contain('/api/v0/ecp1/keypress/Lit_a');
         });
 
+        it('sends a space character as a URI-encoded literal', async () => {
+            const stub = sinon.stub(rokuDeploy as any, 'doPostRequest').resolves({ response: { statusCode: 200 }, body: '' });
+            await rokuDeploy.sendText({ device: rceDevice, text: ' ' });
+            expect(stub.getCall(0).args[0].url).to.contain('/api/v0/ecp1/keypress/Lit_%20');
+        });
+
+        it('sends one keypress per code point (not per UTF-16 code unit) for an astral character like an emoji', async () => {
+            const stub = sinon.stub(rokuDeploy as any, 'doPostRequest').resolves({ response: { statusCode: 200 }, body: '' });
+            await rokuDeploy.sendText({ device: rceDevice, text: '😀' });
+            expect(stub.callCount).to.equal(1);
+            expect(stub.getCall(0).args[0].url).to.equal(`https://device.rce.roku.com/instance/abc/api/v0/ecp1/keypress/${encodeURIComponent('Lit_😀')}`);
+        });
+
         it('routes keydown and keyup through the instance-api route', async () => {
             const stub = sinon.stub(rokuDeploy as any, 'doPostRequest').resolves({ response: { statusCode: 200 }, body: '' });
             await rokuDeploy.keyDown({ device: rceDevice, key: 'Down' });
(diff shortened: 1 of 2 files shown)


TwitchBronBron (approved):
Looks good. Might be worth ensuring our doc blocks explain that the value may not be encoded. Otherwise we will still end up double encoding theirs.

PR #392, opened 2026-09-03
Add sectioned rokudeploy.json config support

### LINKED ISSUE #334: Support roku-deploy.json in v4
We need to add support for `roku-deploy.json` in v4. This will likely require a full restructure of the file, since we have changed almost every option in the system. Things to consider:
- device registry (as seen in #323)
- util function to load the config from projects using this library
- load the config from the cli
- should there be any automatic config loading? 

Haven't totally nailed this down yet, but the general concept is something like:
```js
{
  "logLevel": "info",
    timeout?: number,
    ecpPort?: number;
  "device": {
      host: "1.2.3.4",
      password: "aaaa"
  },
  stage: {
    rootDir: "./myAppFolder",
    out: ".roku-deploy-staging",
    files: [...]
  },
  zip: {
    dir: ".roku-deploy-staging",
    out: "roku-deploy.zip"
  },
  sideload: {
    ...sideloadArgs
  }
```

Then we could also support lifting certain common props like host, password, maybe some of the ports. And might even add ability to read parent props so you can use whatever names you want. We might not need this, but could be nice for flexibility Like maybe:
```js
{
  myOutDir: "somewhere", 
  stage: {
    out: "${myOutDir}"
  }
}
```

Could also support shorthand to keep it more condensed:
```js
{
  "stage.rootDir": "myRootDir",
  "stage.out": "myOutDir,
  "zip.dir": "myOutDir"
}
```
And even more flexible still, we might be able to still auto-connect some of these things depending on your flow:

```js
{ 
  "stage.rootDir": "myRootDir",
  "stage.out": "myOutDir,
  //Use the out from stage step
  "zip.dir": "${stage.out}"
}
```

### LINKED ISSUE #365: `loadConfigFile` exists in v4, but we haven't landed on a known json structure yet
I'm not sure why this function exists quite yet. Considering we haven't decided on a new `rokudeploy.json` schema yet, i don't want to ship with this behavior which then locks us into that. so we either need to solve [#config](https://github.com/rokucommunity/roku-deploy/issues/334) or remove this capability for v4 launch.

Christian-Holbrook:
Implements the config file design from #334 (also resolves #365 by settling the schema).

`rokudeploy.json` (jsonc) = root-level common values + per-command sections, because option names collide across commands (`stage.out` is the staging folder, `zip.out` is the zip path):

```jsonc
{
    "device": "living-room",
    "password": "${ROKU_PASSWORD}",
    "devices": {
        "living-room": { "host": "192.168.1.20" },
        "ci-emulator": { "esn": "X123456789", "rceToken": "${ROKU_RCE_TOKEN}" }
    },
    "stage": { "rootDir": "./src", "out": ".staging" },
    "zip": { "dir": ".staging", "out": "./out/app.zip" }
}
```

Per-command precedence: **CLI args → `[command]` section → root → defaults.**

- New `RokuDeployConfig` type — sections are `Partial<>` of the existing per-command option interfaces, so they can't drift from the method APIs
- `loadConfigFile`: typed return, `${VAR}` env interpolation (clear error naming the variable + json path when unset), warns on invalid registry entries at load
- New `resolveCommandOptions(config, command)` flattens root + section (public, for library consumers)
- CLI: global `--config <path>`; removed yargs `default:`s and `demandOption`s that silently clobbered/blocked file config (validation already exists at the options layer); `rce start`/`stop` accept `--device <registry name>` with token fallback to the entry's `rceToken` (closes the token-env-var question from #372)
- Auto-loading remains CLI-only; the library never reads the filesystem implicitly

Deferred (per issue discussion): dotted-key shorthand, user-defined variables, `${stage.out}` cross-references. JSON schema generation is the follow-up PR.

Closes #334
Closes #365


src/commands/commandUtils.ts
TwitchBronBron, line 1: let's rename `rceCommandUtils.ts` to `commandUtils.ts` and then put this function inside of that file.

src/cli.ts
```diff
@@ -20,6 +20,8 @@ import { RceStopCommand } from './commands/RceStopCommand';
 
 void yargs
 
+    .option('config', { type: 'string', description: 'Path to the config file. Defaults to rokudeploy.json in cwd', global: true })
```
TwitchBronBron: Make sure that, by default, the cli always tries to load `${cwd}/rokudeploy.json`. this `--config` value is for specifying a _different_ path. 

Always log which config file was loaded, even without --verbose. One line: Using config: ./rokudeploy.json. Now nobody's confused, and scripts that care can grep for it or just pass --no-config.

src/cli.ts
```diff
@@ -20,6 +20,8 @@ import { RceStopCommand } from './commands/RceStopCommand';
...
+
     .command('sideload', 'Sideload a zip file or a folder to a remote Roku', (builder) => {
```
TwitchBronBron, line 25: Add a `--no-config` flag to bypass auto-detection entirely. (for all cli actions that allow `--config`)

src/cli.ts
```diff
@@ -67,7 +69,7 @@ void yargs
 
     .command('keyUp', 'send keyup command', (builder) => {
         return builder
-            .option('key', { type: 'string', description: 'The key to send', demandOption: true })
+            .option('key', { type: 'string', description: 'The key to send', demandOption: false })
```
TwitchBronBron: Why is this no longer required? I think it should be. (along with the keyDown and sendText commands

src/RokuDeploy.ts
```diff
@@ -1385,34 +1387,91 @@ export class RokuDeploy {
...
+        }
+        return config;
+    }
+
+    /**
+     * Flatten a loaded config down to the options for one command: root-level common values
+     * overlaid with that command's section (section wins on collision). Section keys for OTHER
+     * commands are stripped. CLI args are merged on top by the caller.
+     */
+    public resolveCommandOptions<T extends Record<string, any> = Record<string, any>>(config: RokuDeployConfig, command: ConfigSectionName): T {
```
TwitchBronBron: Not sure this is the correct name. I know commands are using it, but if we want this to be a public api for consumers, "command" isn't the right term. 

Just a possibility: would this logic make sense to be folded into the `loadConfig` function itself since it seems like it's just applying defaults? just a thought, we can think through it.

src/RokuDeploy.ts
```diff
@@ -1385,34 +1387,91 @@ export class RokuDeploy {
...
+        }
+        Object.assign(result, config[command]);
+        return result as T;
+    }
+
+    /**
+     * Recursively resolve `${VAR}` environment-variable references in every string value.
+     * Throws if a referenced variable is not set, naming the variable and where it was used.
+     */
+    private interpolateEnvVars(node: Record<string, any>, configPath: string, jsonPath: string[]) {
```
TwitchBronBron: feels more like a `util` function rather than on RokuDeploy

src/RokuDeployConfig.ts
```diff
@@ -0,0 +1,199 @@
...
+    sendText?: Partial<SendTextOptions>;
+    /** Options applied only to the `remote-control` command */
+    'remote-control'?: {
+        host?: string;
+        ecpPort?: number;
+    };
+    /** Options applied only to the `rce start` command */
+    'rce.start'?: RceStartConfig;
+    /** Options applied only to the `rce stop` command */
+    'rce.stop'?: RceStopConfig;
```
TwitchBronBron, line 104: I'm not sure _every single_ roku-deploy action belongs here? But maybe? Feels excessive?

src/RokuDeployConfig.ts
```diff
@@ -0,0 +1,199 @@
...
+    devices?: Record<string, DeviceRegistryEntry>;
+    /**
+     * The password for logging in to the developer portal on the target Roku device
+     */
+    password?: string;
+    /**
+     * The username for the roku box. This will always be 'rokudev', but allows to be overridden
+     * just in case roku adds support for custom usernames in the future
+     */
+    username?: string;
```
TwitchBronBron: since this is a new interface, might as well swap their order to be username and then password?

TwitchBronBron (changes requested):

src/cli.ts
Christian-Holbrook: The CLI flow will load the config. This is just to set a different config if you want.

src/cli.ts
Christian-Holbrook: This is now required. I agree that is should be.

src/RokuDeployConfig.ts
```diff
@@ -0,0 +1,171 @@
...
+     */
+    rceToken?: string;
+    /**
+     * The root path to the folder holding your Roku project's source files
+     */
+    rootDir?: string;
+    /**
+     * An array of source file paths, globs, or {src,dest} objects for staging/zipping
+     */
+    files?: FileEntry[];
```
TwitchBronBron, line 71: these don't belong at the top level anymore. they're only applicable to .stage (and maybe .zip?).

src/RokuDeploy.ts
```diff
@@ -1385,20 +1387,30 @@ export class RokuDeploy {
     }
 
     /**
-     * Load options from a rokudeploy.json file. Used by CLI commands to load configuration.
+     * Load a `rokudeploy.json` config file (jsonc: comments and trailing commas allowed).
+     * Resolves `${VAR}` environment-variable references in string values and warns about invalid
```
TwitchBronBron, line 1391: Let's remove the ${VAR} environment variable stuff for now. that's an additive change later, so let's have a deeper discussion about it once v4 lands.

TwitchBronBron (approved):

Issue #77, opened 2021-10-25
Support top-level patterns to external dirs using globstar


TwitchBronBron:
Support using top-level patterns to external directories as long as there's at least 1 globstar. Example:

Consider this pattern:
`"../thirdPartySDK/**/*"`

For this path: `../thirdPartySDK/source/sdk.brs`, its final path would be: `./source/sdk.brs` because the globstar indicates the start of the relative path portion.

If there are multiple globstars, the _first_ globstar is where the relative pattern will begin. 

Here's a unit test that should get you close.

```typescript
it('should support external folders with globstar', async () => {
    //write a file outside of rootDir
    fsExtra.outputFileSync(`${tempDir}/thirdPartySDK/alpha/beta/charlie.brs`, '');
    fsExtra.outputFileSync(`${tempDir}/rootDir/manifest`, '');
    fsExtra.outputFileSync(`${tempDir}/rootDir/source/main.brs`, '');
    await rokuDeploy.stage({
        rootDir: `${tempDir}/rootDir`,
        stagingDir: stagingDir,
        files: [
            '../thirdPartySDK/**/*',
            '**/*'
        ]
    });
    const resultFiles = fastGlob.sync(`**/*`, { cwd: stagingDir });
    expect(resultFiles).to.eql([
        'manifest',
        'source/main.brs',
        'alpha/beta/charlie.brs'
    ]);
});
```

PR #394, opened 2026-09-03
Support top-level patterns to external dirs using globstar

Christian-Holbrook:
A top-level string pattern may now reference files outside rootDir as long as it contains a globstar: the FIRST globstar marks where the dest-relative portion of the path begins (the same rule src;dest; entries already used). External top-level strings without a globstar still throw, with the message updated to mention the globstar escape hatch. getDestPath inherits the same behavior.

Closes #77

### DIFF
--- a/src/RokuDeploy.spec.ts
+++ b/src/RokuDeploy.spec.ts
@@ -4527,7 +4527,7 @@ describe('RokuDeploy', () => {
             expectPathExists(`${stagingDir}/source/main.brs`);
         });
 
-        it('should support external folders with globstar (#77)', async () => {
+        it('should support external folders with globstar', async () => {
             //write a file outside of rootDir
             fsExtra.outputFileSync(`${tempDir}/thirdPartySDK/alpha/beta/charlie.brs`, '');
             writeFiles(rootDir, [
@@ -5973,7 +5973,7 @@ describe('RokuDeploy', () => {
                 ])).to.eql([]);
             });
 
-            it('supports top-level strings to external dirs when they contain a globstar (#77)', async () => {
+            it('supports top-level strings to external dirs when they contain a globstar', async () => {
                 writeFiles(rootDir, [`../thirdPartySDK/source/sdk.brs`]);
                 expect(await resolveFilesArray([
                     '../thirdPartySDK/**/*'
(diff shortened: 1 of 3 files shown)


src/util.spec.ts
```diff
@@ -629,7 +629,7 @@ describe('util', () => {
             expect(s`${destPath}`).to.equal(s`source/main.bs`);
         });
 
-        it('excludes a file found outside the root dir', () => {
+        it('resolves a file found outside the root dir via its globstar (#77)', () => {
```
TwitchBronBron: don't include pr numbers in test cases. 
```suggestion
        it('resolves a file found outside the root dir via its globstar', () => {
```

src/RokuDeploy.spec.ts
```diff
@@ -5952,13 +5973,43 @@ describe('RokuDeploy', () => {
                 ])).to.eql([]);
             });
 
-            it('throws when using top-level string referencing file outside the root dir', async () => {
+            it('supports top-level strings to external dirs when they contain a globstar (#77)', async () => {
```
TwitchBronBron: don't include pr number in test name
```suggestion
            it('supports top-level strings to external dirs when they contain a globstar', async () => {
```

src/RokuDeploy.spec.ts
```diff
@@ -4527,6 +4527,27 @@ describe('RokuDeploy', () => {
             expectPathExists(`${stagingDir}/source/main.brs`);
         });
 
+        it('should support external folders with globstar (#77)', async () => {
```
TwitchBronBron:
```suggestion
        it('should support external folders with globstar', async () => {
```

src/RokuDeploy.spec.ts
```diff
@@ -6106,15 +6157,16 @@ describe('RokuDeploy', () => {
                 }]);
             });
 
-            it('throws exception when top-level strings reference files not under rootDir', async () => {
+            it('resolves top-level strings referencing files not under rootDir via their globstar (#77)', async () => {
```
TwitchBronBron:
```suggestion
            it('resolves top-level strings referencing files not under rootDir via their globstar', async () => {
```

TwitchBronBron (changes requested):

TwitchBronBron (approved):

Issue #404, opened 2026-09-10
DeviceRegistryEntry per-device settings (password, ports, timeout) are silently ignored


Christian-Holbrook:
`DeviceRegistryEntry` documents optional per-device settings:

```ts
export interface DeviceRegistryEntry {
    host?: string;
    esn?: string;
    id?: number;
    instanceUrl?: string;
    rceToken?: string;
    //Optional per-device settings
    password?: string;
    username?: string;
    packagePort?: number;
    ecpPort?: number;
    timeout?: number;
}
```

But `extractDeviceConfig` only extracts the targeting identifier and `rceToken` ‚ the other five fields are read by nothing. So this perfectly reasonable registry:

```jsonc
{
    "devices": {
        "office-tv": { "host": "192.168.0.20", "password": "tv-password" },
        "bedroom":   { "host": "192.168.0.31", "password": "different-password" }
    }
}
```

produces no error and no per-device password ‚ every command uses the ambient top-level `password` (or fails auth), silently.

This gets worse once #334/#392 ships, because the config file feature showcases exactly this registry shape.

## Proposed fix

Wire the settings in: when `resolveDevice` resolves a registry entry, merge the entry's settings into the effective options with precedence:

```
explicit per-call options  >  registry entry settings  >  root/constructor options  >  defaults
```

(That matches how the entry's `rceToken` already behaves for the rce CLI commands.)

Alternative (if we don't want the feature): delete the five fields from the interface. Either change is breaking, which is why this needs to land before v4.0.0.

## Notes

- Needs a precedence test matrix (entry password vs top-level password vs per-call password).
- Docblock on `devices` should state the precedence once implemented.

PR #414, opened 2026-09-11
Fix per-device settings being silently ignored in the devices registry

Christian-Holbrook:
Fixes #404

Per-device settings in the `devices` registry (`password`, `username`, `packagePort`, `ecpPort`, `timeout`) were documented but silently ignored — every call fell back to the ambient/top-level values. They're now honored with the precedence: **per-call option > registry entry > constructor**.

```ts
const rd = new RokuDeploy({ password: 'default', devices: { office: { host: '192.168.0.20', password: 'tv-pass' } } });
await rd.sideload({ device: 'office', ... }); // uses 'tv-pass'
```
- New private mergeOptions() helper replaces the raw { ...this.options, ...options } spread in every public method — the single place where registry-entry settings layer in (raw spreads can't tell per-call values from constructor defaults)
- Settings fields split into a DeviceRegistrySettings interface (DeviceRegistryEntry extends it, consumer shape unchanged); the merge keys derive from Record<keyof DeviceRegistrySettings, true>, so the field list and merge logic can't drift apart
- rceToken/targeting resolution untouched; inline device configs behave exactly as before
- 13-test precedence matrix covering password/ecpPort/timeout from each layer

### DIFF
--- a/src/RokuDeploy.ts
+++ b/src/RokuDeploy.ts
@@ -25,7 +25,7 @@ import type { HttpDetails, RokuDeployError } from './Errors';
 import * as xml2js from 'xml2js';
 import { parse as parseJsonc, printParseErrorCode, type ParseError } from 'jsonc-parser';
 import { util } from './util';
-import type { DeviceRegistryEntry, FileEntry, RokuDeployConstructorOptions, RokuDeployOptions } from './RokuDeployOptions';
+import type { DeviceRegistryEntry, DeviceRegistrySettings, FileEntry, RokuDeployConstructorOptions, RokuDeployOptions } from './RokuDeployOptions';
 import { isLocalDeviceConfig, isRceDeviceConfig, isRceDeviceConfigByEsn, isRceDeviceConfigById, isRceDeviceConfigByUrl, validateDeviceConfig } from './DeviceConfig';
 import type { DeviceConfig, DeviceOption, RceDeviceConfig } from './DeviceConfig';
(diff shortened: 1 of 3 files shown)


TwitchBronBron (approved):

src/RokuDeploy.ts
```diff
@@ -2301,6 +2302,30 @@ export class RokuDeploy {
...
+    /**
+     * Every key of DeviceRegistrySettings, as a record so TypeScript errors if the two ever drift apart.
+     */
+    private static readonly deviceRegistrySettings: Record<keyof DeviceRegistrySettings, true> = { password: true, username: true, packagePort: true, ecpPort: true, timeout: true };
+
+    /**
+     * Merge per-call options over constructor options, layering in any per-device settings from
+     * the targeted devices registry entry. Precedence: per-call > registry entry > constructor.
+     */
+    private mergeOptions<T>(options: T): T {
```
TwitchBronBron, line 2314: It's doing more of that centralized "mergeOptions" function. I don't think I mind having something that helps, but it's super confusing to me how it knows the right sub-config entry to merge together. I'd almost rather see more targeted configs like "getDevices(myConfig, nameOfDevice)" which also merges the settings. Rather than this new config-merger-to-rule-them-all apprach we tried so hard to avoid in earlier v4 commits.

TwitchBronBron (changes requested):

Issue #395, opened 2026-09-03
Support config prop in RokuDeploy constructor


TwitchBronBron:
when https://github.com/rokucommunity/roku-deploy/pull/392 lands, enhance it by supporting a config constructor property. Like this:

```ts
new RokuDeploy({
config: true | 'path-to-config-file'
}
```

When it's true, look at `${cwd}/rokudeploy.json`. When it's a path, resolve using cwd. (if it's absolute path, then cwd is ignored). 

- All of the roku-deploy functions should leverage values from this loaded config if present. 
- maybe add another function for asking roku-deploy to reload that config again? like `rokuDeploy.reloadConfig({ config: true | string})` or something? That's optional, whatever you think.

PR #416, opened 2026-09-16
Support config option in the RokuDeploy constructor

Christian-Holbrook:
Fixes #395

- Adds `config?: boolean | string` to `RokuDeployConstructorOptions`: `true` loads `./rokudeploy.json` from cwd if present (missing file is fine); a string path (resolved against cwd) must exist or construction throws `InvalidOptionError`.
- Precedence is per-call options > constructor options > config file. Only root-level file values are used (command sections are stripped via `loadConfigFile({ section: null })`).
- Adds `reloadConfig()`, which re-reads the same config source with the same rules and rebuilds the instance defaults (no-op when no `config` was given).
- The `config` key itself never leaks into option merging, whether it comes from the constructor or from the file.


src/RokuDeploy.ts
```diff
@@ -86,9 +90,60 @@ export class RokuDeploy {
...
+     * `true` reads `rokudeploy.json` from cwd (a missing file is fine); a string path must exist.
      */
-    private readonly options: RokuDeployConstructorOptions;
+    private loadConstructorConfig(): RootConfigOptions {
+        if (!this.config) {
+            return {};
+        }
+        let configPath: string;
+        if (typeof this.config === 'string') {
+            configPath = path.resolve(process.cwd(), this.config);
```
TwitchBronBron, line 126: Need to think about this. the `process.cwd()` is evaluated on demand here. Should we instead capture the cwd at the time of construction? Or is this correct?

Issue #410, opened 2026-09-10
logLevel option is documented but never read — wire it up


Christian-Holbrook:
`RokuDeployOptions.logLevel` is documented ("The log level. @default LogLevel.log") but nothing ever reads it — setting it has no effect on what roku-deploy logs.

Unlike the other dead fields (see the stagingDir/autoLaunch issue), this one describes behavior we *want*: with rokudeploy.json support landing (#334), `"logLevel": "debug"` in a project config (or `--logLevel` in CI) is a genuinely useful knob for diagnosing deploys without code changes.

## Proposed fix

- When options carry a `logLevel`, apply it to the instance's logger (constructor options and/or per-call — constructor at minimum).
- Consider a `--logLevel` CLI option so `roku-deploy sideload --logLevel debug` works; the config file then supplies it for free via the root section.
- Accept both the string names and numeric levels the `@rokucommunity/logger` types already allow (`LogLevel | LogLevelNumeric`).

## Notes

- This is a bugfix/implementation, not a breaking change — it does not strictly need the v4.0.0 window, but doing it alongside the config-file work is natural since that's what makes the option easily reachable.
- Decide whether per-call `logLevel` overrides should be supported or whether instance-level is enough (per-call requires save/restore semantics around each call — probably not worth it).

PR #418, opened 2026-09-16
Wire up the logLevel option to the logger

Christian-Holbrook:
Fixes #410

- `logLevel` was documented on the options but never read; nothing ever applied it to the logger.
- The constructor `logLevel` now applies to a child logger (the global logger is left untouched), or to a supplied `logger` when one is passed.
- Invalid `logLevel` values throw `InvalidOptionError` via `validateEnum`.
- CLI: a global `--logLevel` flag (with choices) plus `rokudeploy.json` `logLevel` are honored through a yargs middleware that reuses `loadCommandOptions` (so `--config`/`--no-config` behave the same as for commands). The flag wins over the config file.
- Additive; no behavior change unless `logLevel` is set.


src/RokuDeploy.ts
```diff
@@ -47,7 +47,13 @@ export class RokuDeploy {
     constructor(options?: RokuDeployConstructorOptions) {
         this.options = options ?? {};
 
-        this.logger = this.options.logger ?? logger;
+        this.validateEnum(this.options.logLevel, 'logLevel', Object.values(LogLevelNumeric));
+
+        //when a logLevel is set but no custom logger was provided, use a child logger so the shared global logger is not mutated
+        this.logger = this.options.logger ?? (this.options.logLevel === undefined ? logger : logger.createLogger());
```
TwitchBronBron, line 53: This actually flags a bug that I hadn't noticed before. I think every instance of rokuDeploy should get its own new logger. Because consider this:
```ts
const rd1 = new RokuDeploy();
const rd2 = new RokuDeploy();
//this causes rd1's logger to change too since they both used the shared one
rd2.logger.logLevel = 'trace';
```

src/cli.ts
```diff
@@ -17,10 +17,15 @@ import { KeyDownCommand } from './commands/KeyDownCommand';
...
 import { RceStopCommand } from './commands/RceStopCommand';
+import { applyLogLevel } from './commands/commandUtils';
 
 void yargs
 
     .option('config', { type: 'string', description: 'Path to the config file (defaults to rokudeploy.json in cwd). Pass --no-config to skip loading any config file', global: true })
+    .option('logLevel', { type: 'string', description: 'The log level for roku-deploy output. Falls back to "logLevel" in rokudeploy.json', choices: ['off', 'error', 'warn', 'log', 'info', 'debug', 'trace'], global: true })
+
+    //apply --logLevel (or the config file's logLevel) to the logger before running any command
+    .middleware(applyLogLevel)
```
TwitchBronBron, line 28: Is there a reason to have a separate function for this? I think we could probably just inline the applyLogLevel inline here and scrap the function in commandUtils.ts

TwitchBronBron (changes requested):

